PR #17, opened 2026-06-01
Stack-neutral phase personas + project-overridable personas + cross-phase clarity reset

eladrianovalle:
## Summary

Fixes the recurring stale-instruction complaints when running Studio phases outside web-game repos (e.g. Rust + Turbo Genesis), and adds a proper long-term mechanism so personas adapt per-project instead of being overridden inline every run.

Two standing bugs + one new feature, across three commits:

### 1. Stack-neutral defaults (`fd31f25`)
The tech-phase advocate was hardcoded to **"Three.js Technical Architect / WebGL"** (and market to a **"Steam hook"**), so every run in a non-web codebase got a mismatched persona the assistant had to override by hand. Shipped defaults are now stack/platform-neutral. Bridge docs aligned.

### 2. Cross-phase clarity reset (`fd31f25`)
The fresh-run clarity reset (#15) only looked for a prior run in the **same phase folder**, but `clarity.json` is project-level. A prior objective under one phase leaked stale "Clarity-Guided Focus" topics into a new objective under a different phase. Reset now compares the current input against the objective stored in `clarity.json` itself (phase-independent). Adds a cross-phase regression test.

### 3. Project-overridable personas (`0968752`)
Single-phase personas (`market`/`design`/`tech`/`studio`) were hardcoded globals — unlike studio roles, which already support `.studio/roles/*.json`. They are now overridable via **`.studio/personas.toml`**, mirroring the role-override system.

- **`persona_overrides.py`** — load/validate/merge per-phase overrides over `PHASE_DETAILS` (shallow per-phase; `implementer` table merged one level deeper; `integrator` studio-only, `implementer` non-studio-only). `PHASE_DETAILS` is never mutated.
- **`run_phase.py`** — `prepare_run` applies overrides and passes effective details into `build_instruction_doc`.
- **`setup.py`** — new versioned `persona_customization` wizard step (`CURRENT_SETUP_VERSION` 1→2) writes `personas.toml`; optional `suggest_personas_from_stack()` sniffs `Cargo.toml`/`package.json`/`*.csproj` and proposes a stack-appropriate persona.

### Docs (`22e3ec2`)
Surface the Coding Principles section in `CLAUDE.md`, document `.studio/personas.toml`, and update `CLAUDE_CODE_USAGE.md` + the `studio-setup` command.

## Testing
`cd studio && python -m pytest tests/` → **514 passed**. 30 new tests (persona_overrides unit, setup wizard, instruction-generation integration).


eladrianovalle (commented):
APPROVED

Stack-neutrality cleanup is well-scoped: PHASE_DETAILS loses the Three.js/Steam wording, `personas.toml` lets a repo tailor per-phase, and the clarity reset is now phase-independent + self-heals on corrupt JSON. Tests pass (539). Author == reviewer, so leaving this as a self-review comment.

**Must fix**
- None.

**Consider**
- `studio/run_phase.py:1170-1174` — `(artifact_root / ".studio" / "clarity.json").unlink(missing_ok=True)` runs even when `project_clarity is None` on a brand-new project (no file ever existed). It's a no-op, but the condition reads cleaner as `if objective_changed: unlink(...)` and a separate `if project_clarity is None: project_clarity = empty_snapshot(...)`.
- `studio/setup.py:_format_personas_toml` — relies on dict iteration order to emit `[phase]` table keys before the nested `[phase.implementer]` table. That's safe under guaranteed-ordered Python 3.7+ dicts and the validator gates phase keys, but a comment on the ordering invariant would protect future edits.
- `studio/setup.py:apply_persona_customization` / `apply_unstale_config` — when the user later answers "skip" after a prior run wrote `.studio/personas.toml` (or `unstale.toml`), the old file is left in place even though state records the empty choice. Probably fine (skip ≠ revert), but worth documenting in the wizard prose so users aren't surprised.
- `studio/persona_overrides.py:apply_persona_overrides` — `copy.deepcopy(base_fields)` on every phase is cheap here but inconsistent with the "shallow merge" docstring. A `dict(base_fields)` plus a nested copy for `implementer` would match the wording.

PR #19, opened 2026-06-24
feat: contrarian editor mandate + always-on open-questions pre-flight

eladrianovalle:
## Why

Studio runs were overengineering tasks — even the *contrarian* piled on caveats and complexity instead of cutting it, and runs tended to silently assume answers rather than surface what was unsettled. This makes the contrarian an **editor by default** and stops runs from assuming.

### Contrarian = editor, always on
- New shared `CONTRARIAN_MANDATE` (`scopes.py`) injected into **every deliverable-producing contrarian prompt** — single-phase, flat studio, and scoped studio (`generate_scope_prompt`). Bias toward removal / merge / simplification *on top of* flaw-hunting: default to deletion, name the cut specifically, simpler beats more complete, guard the essence.
- The four `PHASE_DETAILS` contrarians reframed as editors (e.g. *Scope-Creep Police & Editor*).
- Appended after the persona line, so it survives `.studio/personas.toml` overrides — genuinely always-on.

### Deliberately OFF in `--mode questions`
- Cutting questions during question-surfacing is exactly backwards. Gated at **both** injection sites: `build_instruction_doc` (`is_qmode`) and `generate_scope_prompt` (new `question_mode` param, threaded through `inject_context` from `output_type`). The scoped path needed its own gate because studio question-mode runs keep scopes active (`_resolve_scopes` only checks `--no-scopes`) — caught in code review.
- `question_mode.py` step 5 now judges question **relevance**, with an explicit guard against dropping genuinely-open questions.

### Open-Questions Pre-Flight (Step 0)
- Every deliverable run now opens by surfacing what's unsettled, pausing on P0 blockers, and recording answers (which raise the clarity score, so later passes only re-ask what's still open) before the iteration loop. Reuses existing decision/clarity machinery — no new module.
- The Decision Point Protocol now states explicitly that every later pass keeps raising new questions.

## Docs
`CLAUDE.md`, `CLAUDE_CODE_USAGE.md`, and both changelogs updated (documentation contract). Also folds in pending `/unstale` doc reconciliation (changelog entries, `INDEX.md` links, setup wizard help strings) that was intermingled in shared files.

## Tests
539 → **546** passing. New coverage: mandate present for contrarian / absent for advocate and question mode at both unit (`test_scopes.py`) and `inject_context` integration level (`test_scoped_debate.py`); pre-flight section renders; question mode omits both mandate and pre-flight (`test_run_phase.py`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,13 @@ All notable changes to TheGameStudio are documented here.
 ## [Unreleased]
 
 ### Added
+- Contrarian editor mandate (always on) — every deliverable run's contrarian now carries a built-in bias toward removal, merge, and simplification on top of flaw-hunting. The advocate piles it on; the contrarian carves out the essence. Shared `CONTRARIAN_MANDATE` injected across single-phase, flat, and scoped studio prompts; phase contrarians reframed as editors. Off in `--mode questions` (there the contrarian judges question relevance, never cuts genuinely-open questions)
+- Open-Questions Pre-Flight (Step 0) — every deliverable run opens by surfacing what is genuinely unsettled, pausing on P0 blockers, and recording answers before the iteration loop, so no run silently assumes; later passes keep raising new questions as they surface. Reuses existing decision/clarity machinery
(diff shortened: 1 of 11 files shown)


eladrianovalle (commented):
APPROVED
The PR keeps the new persona override, contrarian editor mandate, open-question preflight, and stack-agnostic unstale behavior internally consistent, with targeted tests covering the important paths.

Must fix
None.

Consider
None.

PR #21, opened 2026-06-25
feat: Slack/n8n run-digest webhook integration

eladrianovalle:
## What

Adds the first **outbound integration** to Studio: a run-digest notifier that posts a finalized run's status / verdict / summary to a **Slack Incoming Webhook** (Block Kit) and/or an **n8n Webhook node** (flat JSON). Both targets are "HTTP POST a JSON body to a URL", so a single stdlib `urllib` transport serves both — only the payload schema and auth differ.

New module: `studio/integrations/slack_digest.py` (stdlib only, no new deps).

## How it fires

- **`notify` CLI subcommand** is the primary trigger:
  ```bash
  python studio/run_phase.py notify --run-dir <run_dir>            # post to enabled webhooks
  python studio/run_phase.py notify --run-dir <run_dir> --dry-run  # print payloads, no post
  ```
- **Auto-fire on finalize** is wired in the same change but **default-disabled** — a no-op unless `[slack].enabled`/`[n8n].enabled` is set. Soft-fail: a webhook error is logged but never breaks `finalize`.

## Config

`.studio/integrations.toml` (absent → off). Webhook URLs are secrets, resolved from **env vars** named in the config — never committed:

```toml
[slack]
enabled = true
webhook_url_env = "SLACK_WEBHOOK_URL"

[n8n]
enabled = false
webhook_url_env = "N8N_WEBHOOK_URL"
auth_header = "X-API-Key"          # optional (n8n Header Auth)
auth_value_env = "N8N_WEBHOOK_KEY"
```

## Payloads

- **Slack** — Block Kit: header + two-column field section (status/verdict/phase/run id) + context footer, with the required top-level `text` fallback.
- **n8n** — flat JSON addressable as `$json.body.<field>`: `source`, `event`, `phase`, `run_id`, `status`, `verdict`, `iterations_run`, `summary_path`, `summary_text`, `timestamp`.

## Tests

`studio/tests/test_slack_digest.py` — 16 tests with `urllib` mocked (no network): config load/validate, secret resolution, payload builders, success / HTTP error / timeout / 429-retry transport, and orchestration (no targets, missing run.json, dry-run, multi-target post with auth header, skip-without-url).

- **Full suite: 562 passed** (was 546 → +16, zero regressions).

## Docs

- `CLAUDE.md` — CLI command, architecture module entry, config-file entry
- `studio/docs/INTEGRATIONS.md` — setup / usage / security companion doc
- `studio/docs/tickets/TICKET-slack-n8n-digest.md` — the originating ticket + build order

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -135,6 +135,10 @@ python studio/run_phase.py show-metrics --run-dir <path>
 python studio/run_phase.py cleanup --dry-run
 python studio/run_phase.py cleanup
 
+# Outbound notifications (Slack / n8n run digest — see studio/docs/INTEGRATIONS.md)
+python studio/run_phase.py notify --run-dir <run_dir>            # post digest to enabled webhooks
+python studio/run_phase.py notify --run-dir <run_dir> --dry-run  # print payloads without posting
+
 # Cross-repo install (installs slash commands + source into any project)
 python studio/run_phase.py init --target /path/to/project
 python studio/run_phase.py check-install --target /path/to/project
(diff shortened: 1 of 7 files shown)


eladrianovalle (commented):
REJECTED
The integration is generally well-covered, but it currently allows webhook secrets to be stored directly in the repo config.

Must fix
- studio/integrations/slack_digest.py:93: `_resolve_secret()` falls back to literal `webhook_url` / `auth_value`, which breaks the documented "env vars only / never committed" contract and makes it easy to commit a Slack webhook credential. Please remove the literal fallback and require the `*_env` indirection for webhook URLs and auth values.

Consider
- None.

eladrianovalle:
Addressed in 775a615.

`_resolve_secret()` now reads secrets **strictly** from the `*_env` environment-variable indirection — the literal `webhook_url` / `auth_value` fallback is removed entirely. A URL or auth value placed directly in `.studio/integrations.toml` is now ignored (returns `None` → that target is skipped), so a credential can't be committed via config.

Tests updated to lock this in: `test_resolve_secret_no_literal_fallback` (literal in config is ignored) and `test_resolve_secret_no_env_key` (no `*_env` → no secret). Full suite: **563 passing**.

eladrianovalle (commented):
APPROVED
The follow-up removes the committed-config secret fallback and keeps webhook URL and n8n auth resolution env-only.

Must fix
- None.

Consider
- None.

PR #22, opened 2026-06-25
ci: add Phase-1 PR triage gate

eladrianovalle:
Adds the cheap heuristic PR-triage gate (Phase 1 of the agent-driven review pipeline) — same gate as cerebro, stack-tuned for Python. On every PR (and a twice-daily sweep) it labels the PR triage:trivial|small|medium|large|skip by LOC + high-risk paths. No LLM calls, no merges, no comments. Phase 2 (the Multica reviewer) will only spend a Codex review on triage:small PRs. High-risk paths: studio/config/, studio/role_packs/, studio/validators/, studio/studio.manifest.json, studio/pyproject.toml, .github/workflows/, CLAUDE.md.

### DIFF
--- a/.github/workflows/pr-triage.yml
+++ b/.github/workflows/pr-triage.yml
@@ -1,0 +1,130 @@
+name: PR triage (heuristic)
+
+# Phase 1 of the agent-driven PR review pipeline. This workflow is the cheap
+# gate: every open non-draft PR gets one of the `triage:*` labels based on
+# size + which paths it touches. NO LLM calls; NO merges; NO comments. The
+# label is the audit trail. Phase 2 (the Multica review agent) consumes the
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
APPROVED
The workflow cleanly adds a cheap deterministic PR triage gate and the existing label/check setup matches its assumptions.

Must fix
None.

Consider
Forked PRs may not be labelable from the `pull_request` event with the default token, but the scheduled sweep should still catch anything that matters.

eladrianovalle (commented):
APPROVED
The new triage workflow is behaving as intended on this PR, including applying the high-risk `triage:medium` label and completing its own check.

Must fix
None.

Consider
The `labeled`/`unlabeled` triggers will cause one converging follow-up run after the workflow edits labels; that is acceptable, but it may be worth removing those triggers later if Actions noise becomes annoying.

PR #25, opened 2026-06-26
Fix installer omitting integrations/ package (breaks every command in installed projects)

eladrianovalle:
## Problem

`studio/run_phase.py` imports `integrations.slack_digest` at module top level:

```python
from integrations.slack_digest import load_integrations_config, notify_run
```

…but `install.py`'s `SOURCE_FILES` never listed the `integrations/` package. So a cross-repo install (`init`/`update`) copies `run_phase.py` without its dependency, and the installed `.studio/source/run_phase.py` crashes on import with:

```
ModuleNotFoundError: No module named 'integrations'
```

Because the import is at the top level, this breaks **every** subcommand in an installed project (not just `notify`) — so all `/run-phase` and `/run-studio-phase` slash commands are unusable after `init`.

## Fix

Add the two `integrations/` files to `SOURCE_FILES`:

```
integrations/__init__.py
integrations/slack_digest.py
```

## Verification

- Reproduced: a fresh `init --target <dir>` produced a `.studio/source/` with no `integrations/`, and `python .studio/source/run_phase.py show-clarity` crashed on import.
- After the fix, a throwaway `init --target` copies `integrations/` and `show-clarity` (and `notify --help`) run cleanly with no import error.

### DIFF
--- a/studio/install.py
+++ b/studio/install.py
@@ -44,6 +44,8 @@ SOURCE_FILES = [
     "validators/__init__.py",
     "validators/document_validator.py",
     "validators/code_validator.py",
+    "integrations/__init__.py",
+    "integrations/slack_digest.py",
     "config/scopes.toml",
     "config/studio_settings.toml",
     "docs/STUDIO_BRIDGE_TEMPLATE.md",


eladrianovalle (commented):
APPROVED

Correctly fixes a real install-time crash by adding the missing `integrations/` package to `SOURCE_FILES`.

**Verified**
- `run_phase.py:73` imports `from integrations.slack_digest import ...` at module top level, so the missing package broke every subcommand in an installed project, not just `notify`.
- `integrations/__init__.py` and `integrations/slack_digest.py` both exist, so `_collect_source_files` (which only includes files that `exist()`) will now pick them up.
- The copy loop does `dst.parent.mkdir(parents=True, exist_ok=True)`, so the new `integrations/` subdir is created on install — same pattern already proven by the `validators/` entries.
- No existing glob covered `integrations/`, so the omission was genuine.

Minimal, surgical, and consistent with the surrounding entries. No blockers.

PR #26, opened 2026-06-26
Slack digest: include a readable body (prefer digest.md) + final-doc pointer

eladrianovalle:
Two commits, both about making the Slack digest actually useful.

## 1. Include the run body + a final-doc pointer
`build_slack_blocks(meta)` only sent header + status/verdict/phase/run-id + footer — it never read the run's summary, so the message carried no content and no way to find the doc. Now it appends the body (converted to Slack `mrkdwn`, truncated to 2700 chars) and a `Final doc:` pointer.

## 2. Prefer a plain-language `digest.md` over `summary.md`
`summary.md` is often dense/technical and gets truncated. If a run authors a short, plain-language `digest.md` (or `summary_human.md`), the Slack body uses **that**; otherwise it falls back to `summary.md`. The `Final doc:` pointer always targets the full `summary.md`.

`_md_to_slack()` handles markdown→Slack conversion (headings→bold, `-`/`*` bullets→`•`, `**x**`→`*x*`). n8n payload unchanged. Soft-fail behavior unchanged.

## Verification
- `notify --dry-run` on a real finalized run: blocks `[header, fields, divider, body(digest.md), final-doc, divider, footer]`, plain version, no truncation.
- Live sends returned `slack: ok (HTTP 200)` for both the technical-summary and the digest.md paths.

### DIFF
--- a/studio/integrations/slack_digest.py
+++ b/studio/integrations/slack_digest.py
@@ -36,6 +36,7 @@ except ModuleNotFoundError:
 
 import json
 import os
+import re
 import time
 import urllib.error
 import urllib.request
@@ -45,6 +46,9 @@ from typing import Dict, List, Optional, Tuple
 
 INTEGRATIONS_FILENAME = "integrations.toml"
 _SUMMARY_MAX_CHARS = 600
+# Slack section text objects cap at 3000 chars; leave headroom for the
+# truncation notice appended below.
+_SLACK_SUMMARY_MAX_CHARS = 2700
 _USER_AGENT = "TheGameStudio-digest/1.0"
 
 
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
APPROVED

Cleanly wires the run `summary.md` (converted to Slack mrkdwn and truncated) plus a final-doc pointer into the Slack digest, leaving the n8n payload and soft-fail behavior untouched.

**Must fix**
- None.

**Consider**
- `_summary_doc_path` always returns a path, so the `*Final doc:*` pointer is appended even when `summary.md` is absent (summary_text == "") — the digest would point at a file that doesn't exist. Gating the pointer on the summary actually being present (or on the file existing) would avoid a dangling reference.
- `_md_to_slack` truncates after the `**`→`*` collapse, so a cut can land mid-`*bold*` span and leave an unbalanced `*`. Cosmetic only (Slack renders the stray star literally), but trimming to the last newline before the cap would read cleaner.
- `_read_summary_text` and `_summary_doc_path` now duplicate the `summary_path` resolution; a small shared helper would keep them from drifting.

eladrianovalle (commented):
APPROVED

Wires the run summary + a final-doc pointer into the Slack digest with a focused, well-documented markdown→mrkdwn conversion; n8n payload and soft-fail behavior are untouched, and char budgeting (2700 + notice < Slack's 3000 section cap) is correct.

**Must fix**
- None.

**Consider**
- `notify_run`: `_summary_doc_path()` always returns a resolved path, so the `Final doc:` section renders even when `summary.md` is absent (`_read_summary_text` returned ""). The digest would then point at a file that doesn't exist. Gating `doc_path` on the summary actually being found would avoid the dangling pointer.
- `notify_run`: when both targets are enabled, `_read_summary_text(run_dir, meta)` runs once for Slack and again for n8n. Hoisting it to a single read is a trivial cleanup.
- `_md_to_slack`: the global `text.replace("**", "*")` runs after heading wrapping, so a markdown heading that itself contains bold (e.g. `## **x**`) collapses to a stray double-star (`**x**`) in Slack. Edge case, non-blocking.

PR #28, opened 2026-06-27
Fix check-install/update false 'up to date' via stale snapshot (#20)

### LINKED ISSUE #20: check-install reports false 'up to date' when run via installed snapshot
## Summary

`/studio-update` (and `check-install` in general) can report **"up to date"** even when the installed project is many commits behind upstream. This silently blocks updates.

## Root cause

The bundled `/studio-update` slash command invokes `check-install` via the **installed snapshot** at `<project>/.studio/source/run_phase.py`. That snapshot compares the installed files against *itself*, not against the live upstream source. So:

- If the installed snapshot is stale, it never detects that it's stale.
- The comparison also depends on whatever branch the upstream `source_path` repo happens to be checked out on. If upstream is on a feature branch, or local `main` is behind `origin/main`, the diff is computed against the wrong tree.

## Reproduction

1. Install Studio into a project (records e.g. commit `8ce5584`).
2. Land several commits on upstream `main`.
3. Leave the upstream repo checked out on a feature branch, or with local `main` behind `origin/main`.
4. Run `/studio-update` → `python .studio/source/run_phase.py check-install --target .`
5. Observed: `Studio at <path> is up to date.`
6. Run the **same** check from the upstream working copy instead:
   `python studio/run_phase.py check-install --target <project>`
7. Actual upstream state: `needs updating` — 6 changed, 2 missing files.

(Real-world: installed `8ce5584` (Apr 14) reported up-to-date against upstream `5c24c36` (Jun 24); upstream invocation correctly flagged the diff.)

## Suggested fixes (any one or combination)

- **check-install should compare against the live source repo**, resolving `source_path` from `VERSION` and reading current files there — not against the installed `.studio/source/` snapshot.
- **Guard on branch/sync state**: if the `source_path` repo is not on `main`, or local `main` is behind its upstream, warn loudly (or fetch + compare against `origin/main`) instead of silently reporting up-to-date.
- **Re-exec the upstream `run_phase.py`** for `check-install`/`update` rather than the installed copy, so version-skew in the snapshot can't mask updates.

## Workaround

Run check/update directly from the upstream checkout, with `main` current:

```
cd <studio-repo> && git checkout main && git pull
python studio/run_phase.py check-install --target <project>
python studio/run_phase.py update --target <project>
```

eladrianovalle:
Fixes #20.

## Problem
`/studio-update` invokes `check-install` through the **installed snapshot** at `<project>/.studio/source/run_phase.py`. The default source root resolved to that snapshot, so `check_studio` compared the snapshot against its own `MANIFEST.json` — always reporting **"up to date"**, silently blocking updates. `update_studio` shared the flaw: it re-installed from the snapshot onto itself, so it could never actually pull upstream changes.

Real-world: an install at commit `8ce5584` reported up-to-date against upstream `5c24c36` (6 changed, 2 missing files).

## Fix
New `_resolve_source_dir(target, studio_dir)`:
- Explicit `studio_dir` (tests / upstream invocation) → used as-is, unchanged.
- Default root that **is** the installed snapshot → resolve the live upstream from `VERSION.source_path` and compare/copy against that.
- Live source unresolvable (source_path missing, moved, or self-referential) → return a loud **WARNING** instead of a silent, unreliable "up to date".

Warning is threaded through `check_studio`/`update_studio` result dicts and printed by the `check-install`/`update` CLI handlers.

## Verification
- 3 regression tests (stale-snapshot detection, update-refreshes-from-live, warning-on-unresolvable). Full suite: **580 passed**.
- End-to-end via the snapshot CLI path: stale install now reports `Changed: docs/CODING_PRINCIPLES.md`, `update` restores it, and a broken `source_path` prints the WARNING.

### DIFF
--- a/studio/install.py
+++ b/studio/install.py
@@ -22,7 +22,7 @@ import shutil
 import subprocess
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Tuple
 
 # Files to copy from studio/ into .studio/source/
 SOURCE_FILES = [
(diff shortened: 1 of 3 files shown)


eladrianovalle (commented):
APPROVED

Correctly fixes the snapshot-vs-self comparison so check-install/update resolve the live upstream source from VERSION.source_path instead of silently reporting "up to date".

**Must fix**
- None.

**Consider**
- `update_studio` calls `_resolve_source_dir` then `check_studio(target, source_dir)`, which resolves again (trivially, since source_dir is now non-None). Harmless but a touch redundant.
- When the live source is unresolvable, `source_dir` falls back to the snapshot, so `check_studio` will still report "up to date" alongside the WARNING. Acceptable degradation given the loud warning, but worth a comment noting the result is intentionally unreliable in that path.

eladrianovalle (commented):
APPROVED

Re-review after rebase: the head moved from 872a761 to f2aeaac, but this is purely a rebase onto current main (it picked up the slack-digest merges #26/#27). The PR's own change (`install.py`, `run_phase.py`, `test_install.py`) is byte-for-byte identical to the revision I already approved.

**Must fix**
- None.

**Consider**
- Same two non-blocking notes from the prior review still stand (redundant re-resolve in `update_studio`; "up to date" still reported alongside the WARNING when the live source is unresolvable). Neither blocks.

PR #30, opened 2026-06-28
feat: implementation writer/editor loop (spec + workflow + /studio-implement)

eladrianovalle:
## What

Brings the advocate/contrarian (writer/editor) cadence into **implementation**. A writer agent builds one complete MVI unit and commits its passing state; a fresh editor agent (reusing `CONTRARIAN_MANDATE`) cuts/refines it against the writer's diff, reverting if an edit breaks green; then the unit is delivered.

**Pipeline, not debate** — the editor applies edits directly (no ping-pong); the only branch is a one-shot revert.

## Files

| File | Role |
|---|---|
| `studio/docs/IMPLEMENTATION_LOOP_SPEC.md` | The design spec |
| `.claude/workflows/implementation-loop.js` | Orchestration shell (built, **not yet run**) |
| `.claude/commands/studio-implement.md` | `/studio-implement` — the authorized trigger |
| `.gitignore` | Track `.claude/workflows/` so the workflow ships with its command |

## Design highlights

- **Gate split** into an *entry* moment (writer **declares done** + tests green + ruff → triggers the editor) and an *exit* moment (editor's authoritative `mvi_verdict` can overturn the writer's claim). This resolves the circularity of "who defines MVI."
- **In-memory handoff anchored to `writer_sha`** — the commit of the writer's passing state serves as the diff base, the unit boundary, and the revert restore-point.
- **Portable core:** prompts, gate criteria, cadence params, and handoff shape are data; a port to another runtime (LangGraph / ADK / OpenAI) rewrites only the ~40-line shell.
- Gate reuses the existing `CodeValidator`; mutation/MVI checks are honestly labeled **agent-attested**, not machine-enforced.

Refined once via a **Product + Engineering** role pass (see the spec's Refinement log).

## Status / scope

- **Spec + shell + trigger only.** The loop has **not been run**. First run (default target: building its own config loader, `studio/impl_loop.py`) is also the first cadence-lab data point.
- Future build phases (in the spec): runnable loop on one unit → extract `implementation_loop.toml` + `impl_loop.py` → cadence lab.
- Command is local to this repo (not wired into the cross-repo installer yet).

## Note for reviewers

Running `/studio-implement` spawns multiple agents, edits files in place, and makes a real git commit — that's why it's gated behind an explicit command. It refuses to run on a dirty tree and branches off `main` automatically.

### DIFF
--- a/.claude/commands/studio-implement.md
+++ b/.claude/commands/studio-implement.md
@@ -1,0 +1,102 @@
+# Studio Implement (Writer/Editor Loop)
+
+Build one MVI unit through the implementation writer/editor loop: a writer agent builds a
+complete, usable unit and commits its passing state; a fresh editor agent (carrying the
+`CONTRARIAN_MANDATE`) cuts/refines it, reverting if an edit breaks green; then it's delivered.
+See `studio/docs/IMPLEMENTATION_LOOP_SPEC.md` for the full design.
+
+This command is the **authorized, unequivocal trigger** for the loop. Invoking it is explicit
+opt-in to run the `implementation-loop` workflow (which spawns multiple agents). When the user
+asks in natural language to "implement X with the loop" / "run the writer-editor loop on X",
+treat that as an invocation of this command.
+
+## Arguments
+
+- `$ARGUMENTS` — Required. Free-text description of the ONE unit to build. Should describe a
+  complete, usable interaction (MVI), not a partial component.
+- Optional `--branch <name>` — branch to run on (default: derive `impl/<unit_id>`).
+- Optional `--test "<cmd>"` — override the inferred test command.
+- Optional `--plan` — echo the parsed plan and STOP (dry run); do not run the loop.
+
+## Instructions
+
+Run behavior is **echo-plan-then-run**: print the plan, then immediately run the loop (no
+separate confirmation step — the slash invocation is the authorization). `--plan` is the only
+thing that stops before running.
+
+### Step 1 — Parse the request into a unit object
+
+From `$ARGUMENTS`, construct the workflow `args`:
+
+- `unit_id` — a short snake_case slug derived from the request (e.g. `profile_create_view`).
+- `title` — a one-line MVI-framed statement of the usable outcome ("X can do Y …"). If the
+  request describes a partial component ("add a data model", "define the schema"), reframe it as
+  the smallest usable interaction, or stop and say it isn't an MVI unit.
+- `instructions` — concrete build steps: what files to create/modify, what tests to write. Keep
+  it to this one unit; no speculative scope.
+- `test_command` — infer from the repo's stack unless `--test` is given. Detect from marker files:
+  - Python (`pyproject.toml`/`pytest`): `python -m pytest <path> -q` (in this repo: `cd studio && python -m pytest tests/<file> -q`).
+  - Node (`package.json`): `npm test` or the project's test script.
+  - Rust (`Cargo.toml`): `cargo test`.
+  - Unity / other: ask if you cannot infer a runnable command.
+- `static_check` — `ruff check <path>` for Python; the project linter otherwise; omit if none.
+
+### Step 2 — Pick the target branch (safety)
+
+The loop edits files in place and makes a real `git commit`, and the writer + editor stages must
+share one working tree. So:
+
+```bash
+git rev-parse --abbrev-ref HEAD
+```
+
+- If `--branch` was given, check it out (create if missing).
+- Else if on `main`/`master`, create and switch to `impl/<unit_id>`.
+- Else use the current branch (it's already a feature branch).
+
+If the working tree is dirty, stop and tell the user to commit/stash first — the loop's revert
+relies on a clean base.
+
+### Step 3 — Echo the plan
+
+Print a short block, then proceed (unless `--plan`):
+
+```
+/studio-implement
+  unit_id:  <slug>
+  title:    <MVI outcome>
+  branch:   <branch>
+  tests:    <test_command>
+  builds:   <1-line of what files/behavior>
+```
+
+If `--plan`, stop here.
+
+### Step 4 — Run the loop
+
+Call the **Workflow** tool with the `implementation-loop` workflow, passing the parsed unit as
+`args`:
+
+```
+Workflow({ name: "implementation-loop", args: {
+  unit_id, title, instructions, test_command, static_check
+}})
+```
+
+The workflow runs in the background and notifies on completion. Do not re-run it or poll;
+wait for the completion notification.
+
+### Step 5 — Report the result
+
+The workflow returns `{ delivered, flagged, editorRan, finalVersion, writer, editor }`. Summarize
+for the user:
+
+- **What was built** + final test status (`editor.tests` / `writer.tests`).
+- **What the editor changed** (`editor.edits`) — the cuts/merges/restructures, or "nothing to cut".
+- **MVI verdict** (`editor.mvi_verdict`) — and call out explicitly if the editor *overturned* the
+  writer's `mvi_claimed` (unit not actually usable → it's flagged, not silently shipped).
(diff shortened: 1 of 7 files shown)


eladrianovalle (commented):
APPROVED

A clean spec + unrun orchestration shell for the writer/editor implementation loop; the workflow logic is internally consistent and the scope (spec + shell + trigger, not run) is honestly stated.

**Must fix**
- None.

**Consider**
- `implementation-loop.js:267` — `(args && args.unit_id) ? args : DEFAULT_UNIT` silently falls back to building the config loader if `args` is passed without `unit_id`. The slash command always sets `unit_id`, so this only bites a direct Workflow call, but a thrown error on malformed `args` would fail louder than silently building the wrong unit.
- `implementation-loop.js:273` — entry gate uses `writer.static_ok !== false`, so an absent `static_ok` (it's optional in `WRITER_HANDOFF`) passes the gate. Fine as a lenient default, but worth deciding whether "static check not reported" should really admit to the editor.
- Gate-fail return at `:281` omits `finalVersion`/`editor`, while the success return and the command's Step 5 reporting expect the fuller shape. Reporting tolerates the missing keys, just noting the asymmetry.
- The spec's delivery-time `runFullSuite()` (SPEC §"executor shell") isn't in the shell yet — correctly deferred to Build Phase 1, flagging so it isn't lost.

eladrianovalle (commented):
APPROVED

Spec + thin disposable executor for the writer/editor implementation loop — explicitly not-yet-run, and the gate logic is honest about what's machine-enforced vs agent-attested. The only executable code (`implementation-loop.js`) is orchestration-only and reads cleanly; entry/exit gates and the one-shot revert match the spec.

**Must fix**
- (none)

**Consider**
- `.claude/workflows/implementation-loop.js:229,253` — `writerPrompt`/`editorPrompt` interpolate `u.static_check` / `u.test_command` directly. The command (`studio-implement.md:48`) says to omit `static_check` when there's no linter; an omitted field renders the prompt line as `` the static check: `undefined` ``. Guard the line or default it.
- `implementation-loop.js:292-300` — the forced-revert safety net resets to `writer_sha` via a side agent but never updates the returned `editor.reverted` / `editor.tests`, so the result reports `reverted=false` and `flagged` on a tree that's actually back at the writer's green state. Cosmetic in the payload, but the return should reflect `reverted=true`.
- The shell trusts the command's dirty-tree/branch guard (`studio-implement.md` Step 2). Invoked directly via the Workflow tool it'll `git commit` on whatever HEAD is checked out. Acceptable given the gating, but a one-line clean-tree assert at the top of the shell would make it safe standalone.

eladrianovalle (commented):
APPROVED

Follow-up push (`6247a60..d1aeaac`) touches only the orchestration shell and cleanly resolves two items from my prior pass — no new blockers.

**Must fix**
- None.

**Consider**
- The forced-revert safety net (`implementation-loop.js:187`) still doesn't update `editor.reverted`, so the closing `log` and the returned `editor` object report `reverted=false`/`committed=false` on a tree that's actually back at `writer_sha`. Cosmetic and pre-existing, but the payload still misrepresents that path.
- `git commit -am` in the commit safety net (`:198`) only stages tracked, modified files — if an editor edit ever adds a new file, it'd be left uncommitted. Fine for the current "refine existing files" editor contract; worth a note if editors are ever allowed to create files.

Both changes are good improvements: `{ ...DEFAULT_UNIT, ...overrides }` makes partial `args` safe and the `committed` field + safety net close the keep-but-didn't-commit gap.

eladrianovalle (commented):
APPROVED

Follow-up push (`d1aeaac..c781b69c`) is a 7-line payload-accuracy fix to the orchestration shell — no behavior change, no new blockers.

**Must fix**
- None.

**Consider**
- `implementation-loop.js:198` — the safety-net path sets `editor.committed = true` unconditionally, but the agent prompt itself allows for "git reports nothing to commit." If `editHeld` is true yet there's genuinely nothing staged, `committed` now reports `true` on a no-op commit. Narrow edge given `editHeld` implies kept edits; cosmetic only.

Both edits cleanly resolve the `reverted`/`committed` misreporting I flagged last pass, and the `git commit -am` note documents the intentional tracked-only contract well.

eladrianovalle (commented):
APPROVED

Follow-up: merging #31 into the base advanced this branch (`c781b69c..29cdac14`), so the net-new content here is the impl_loop config loader — `impl_loop.py`, the shipped `implementation_loop.toml`, and 14 tests. It's clean and faithfully mirrors the `scopes.py` loader pattern (tomllib/tomli fallback, explicit → `.studio/` → shipped → defaults chain, explicit-missing-path raises while end-of-chain miss yields defaults). `__post_init__` validation is thorough and the tests cover the cases that matter (overrides, shallow-merge inheritance, `.studio` precedence, malformed TOML). No blockers.

**Must fix**
- None.

**Consider**
- `impl_loop.py:53` — `static_checks` is validated as a list but its elements aren't type-checked, so a TOML `static_checks = [1, 2]` passes `__post_init__` and only fails later when something tries to run a non-string check. A one-line `all(isinstance(c, str) ...)` guard would fail at load time instead. Non-blocking.
- `impl_loop.py:51` — `output_budget` rejects non-ints and bools but accepts `0`/negatives; a `>= 0` (or `> 0`) check would reject a nonsensical word cap up front. Minor.

eladrianovalle (commented):
APPROVED

Follow-up push (`c781b69c..29cdac14`, merge of #31) lands the first real loop output — `impl_loop.py`, its shipped `implementation_loop.toml`, and a 15-case test suite — and it's clean: a faithful port of the `scopes.py` / `load_scopes_config()` pattern with honest resolution-chain semantics (explicit-path miss raises, end-of-chain miss returns defaults).

**Must fix**
- None.

**Consider**
- `studio/tests/test_impl_loop.py:8` — `from impl_loop import …` (bare module) relies on `studio/` already being on `sys.path`, same as `test_scopes.py`. Fine if the gate's `pytest -q` runs from a cwd/conftest where that resolves; just flagging that CI greenness here is invocation-dir dependent, since the import isn't package-qualified.
- `impl_loop.py:62` — `static_checks` is validated as a list but its elements aren't type-checked, so `static_checks = [1, 2]` passes `__post_init__` and feeds non-string lint names downstream. Mirrors the `output_budget`/bool rigor elsewhere; cheap to tighten if you want element validation.
- `impl_loop.py:67` — `output_budget` rejects non-ints and bools but not negatives/zero; a `0` or `-1` word cap would validate. Cosmetic given it's an attested rationale budget, not a hard limit.

eladrianovalle (commented):
APPROVED

Follow-up push (`c781b69..29cdac14`) merges PR #31, delivering the loop's first real output — the `impl_loop.py` config loader (`LoopConfig` + `load_loop_config`), its shipped `implementation_loop.toml`, and 16 tests. It cleanly mirrors the `scopes.py` / `load_scopes_config()` pattern, the resolution chain matches SPEC §4, and the validation is honest about what's enforced. No new blockers.

**Must fix**
- None.

**Consider**
- `impl_loop.py:140-148` — the table merge reads only known keys, so an unknown or typo'd key (`output_budgett`, or a key under the wrong table) is silently dropped. Given commit `7f00c8d5` deliberately made a bad `read_scope` and a missing explicit path *loud*, rejecting unknown keys/tables would be consistent with that same "typos are errors" stance. Non-blocking, but worth a deliberate decision rather than leaving it lenient by omission.
- `impl_loop.py:50` — `static_checks` is validated as a list but its elements aren't type-checked, so `static_checks = [1, 2]` in TOML passes the gate and only fails later when something tries to run a non-string check. A one-line `all(isinstance(c, str) ...)` would fail it at load time, matching the loudness elsewhere.
- `test_impl_loop.py` imports `from impl_loop import ...` (bare module), which relies on the tests running with `studio/` on `sys.path`. Fine if that's the established convention for this package's suite; just flagging the implicit path assumption since the rest of the loader is careful about resolution.

PR #31, opened 2026-06-28
feat: impl_loop config loader (LoopConfig + load_loop_config) — first writer/editor loop output

eladrianovalle:
## What

The implementation writer/editor loop's **Phase 2**: a config loader for `implementation_loop.toml`. Notably, this unit was **built by the loop itself** (`/studio-implement`) — the first real end-to-end run, dogfooded on Studio.

Stacked on #30 (base: `feat/implementation-loop`).

## Files

- `studio/impl_loop.py` — `LoopConfig` dataclass + `load_loop_config()`, mirroring `scopes.py` (tomllib/tomli fallback, resolution chain explicit → `.studio/` → shipped → defaults, `__post_init__` validation, `editor_enabled` property).
- `studio/config/implementation_loop.toml` — shipped default (spec §4 tables).
- `studio/tests/test_impl_loop.py` — **14 tests**: valid parse, missing→defaults, malformed TOML→clear error, `.studio` override beats shipped default.

## How it was produced (loop trace)

| Stage | Result |
|---|---|
| **Writer** | Built the unit; 14 tests green; mutation check (3 assertions broken, all caught); flagged 3 `load_bearing` items (e.g. the `bool`-subclasses-`int` guard on `output_budget`). Committed at `e346a53`. |
| **Editor** (CONTRARIAN_MANDATE) | One substantive cut — collapsed the TOML schema that was triplicated across docstring/config/spec down to a pointer (the drift `/unstale` guards against). `mvi_verdict: true`, nothing reverted. Respected all 3 load_bearing flags. Committed at `50e02b0`. |

## Cadence-lab data point (first run)

- 2 agents, ~88k tokens, ~4.5 min. Yield: 1 substantive cut on a clean ~130-line unit — clears the success metric, but thinly. Signal: editor yield is low on small well-written units; expect more value on larger/messier ones.
- Two loop bugs this run surfaced (args plumbing + editor not committing) are fixed in #30 (`d1aeaac`). They were worked around manually for this unit.

### DIFF
--- a/studio/config/implementation_loop.toml
+++ b/studio/config/implementation_loop.toml
@@ -1,0 +1,17 @@
+# Default configuration for the implementation writer/editor loop.
+#
+# Follows the scopes.toml pattern: shipped default, overridable by
+# .studio/implementation_loop.toml. See studio/docs/IMPLEMENTATION_LOOP_SPEC.md §4.
+
+[loop]
+deliver_on_gate_fail = true   # if the writer can't reach green, deliver flagged (uncommitted) rather than spin
+
+[gate]
+test_command = "pytest -q"    # per-repo; the only stack-specific knob. Runs the unit-scoped tests.
+static_checks = ["ruff"]      # CodeValidator static/lint half (mypy opt-in, off by default)
+require_mutation_check = true # records the attested mutation signal; not machine-enforced yet
+
+[editor]
+mandate = "contrarian"        # reuse CONTRARIAN_MANDATE; "off" disables the editor pass
+read_scope = "touched+importers"  # bound the editor's reads: diff + files_touched + direct importers
+output_budget = 400           # word cap on the editor's rationale (matches scope output budgets)
--- a/studio/impl_loop.py
+++ b/studio/impl_loop.py
@@ -84,21 +84,9 @@ def load_loop_config(path: Path | None = None, studio_root: Path | None = None) 
     explicit or end-of-chain) yields the default LoopConfig rather than an error —
     the loop ships with a working default, so absence is not a failure.
 
-    Expected format (all tables/keys optional; unspecified keys inherit defaults):
-    ```toml
-    [loop]
-    deliver_on_gate_fail = true
-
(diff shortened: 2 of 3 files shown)


eladrianovalle (commented):
APPROVED

A clean, well-tested config loader for the implementation loop that faithfully mirrors the `scopes.py` pattern; validation lives in `__post_init__`, the resolution chain is explicit, and the 14 tests cover the meaningful branches (defaults, partial-merge, override precedence, malformed TOML, missing file).

**Must fix**
- None.

**Consider**
- `impl_loop.py:104` — an explicit `path` that doesn't exist silently returns defaults (via the `not config_path.exists()` guard) rather than raising. It's documented and tested as intentional ("unlike scopes"), but a caller that passes a typo'd `--config` path gets defaults with no signal. Worth deciding whether an *explicit* missing path should be louder than an end-of-chain one.
- `impl_loop.py:120-128` — only `mandate`, `static_checks`, and `output_budget` are type/value-validated; `test_command`, `read_scope`, `deliver_on_gate_fail`, `require_mutation_check` pass through whatever TOML yields. Fine for now since TOML constrains the shapes, but `read_scope` in particular is a free string with a fixed set of valid values upstream.
- Self-approval is blocked since the PR author is me, so this is a `--comment` review — treat it as an approval for merge-gate purposes.

eladrianovalle (commented):
APPROVED

Follow-up push (`50e02b0..7f00c8d`) cleanly resolves both items from my prior pass — no new blockers.

**Must fix**
- None.

**Consider**
- None outstanding. An explicit missing `path` now raises `FileNotFoundError` (end-of-chain absence still yields defaults), and `__post_init__` now validates the full field set — `read_scope` against `VALID_READ_SCOPES`, `test_command` as `str`, and the two bool knobs — so a stray TOML value fails loudly. The two added tests cover both paths.

Self-approval is blocked since the PR author is me, so this is a `--comment` review — treat it as an approval for merge-gate purposes.

PR #32, opened 2026-06-28
fix(workflow): parse args when forwarded as a JSON string

eladrianovalle:
## Root cause

The implementation-loop workflow silently ignored custom units passed via `/studio-implement` — both real runs fell back to `DEFAULT_UNIT` (building the already-merged config loader, doing nothing).

A no-agent probe nailed it: the runtime forwards the Workflow `args` to the script as a **JSON-encoded string**, not a parsed object —

```json
{"received":"{\"hello\": \"world\", ...}","type":"string"}
```

So `args.unit_id` was `undefined` and the `typeof args === 'object'` merge (from an earlier fix attempt) rejected the string → fallback to defaults every time.

## Fix

`JSON.parse` when `args` is a string (fallback to `{}` on parse failure), then merge over `DEFAULT_UNIT` as before. Custom units now actually drive the loop.

This is the real fix for the args plumbing flagged across #30/#31. Verified the failure mode and the string→object normalization; the next loop run with explicit args will confirm end-to-end (the `Unit:` log line will show the passed unit_id instead of the default).

### DIFF
--- a/.claude/workflows/implementation-loop.js
+++ b/.claude/workflows/implementation-loop.js
@@ -157,7 +157,13 @@ function editorPrompt(u, writer) {
 // Merge any provided args over the default unit field-by-field: provided keys win,
 // unspecified keys fall back to DEFAULT_UNIT (so partial args are safe). The log
 // makes the resolved unit visible — if args didn't arrive, this shows the default.
-const overrides = (args && typeof args === 'object' && !Array.isArray(args)) ? args : {}
+// NOTE: args arrives as a JSON-encoded STRING (the runtime serializes it when forwarding),
+// not a parsed object — normalize before merging, else everything falls back to DEFAULT_UNIT.
+let parsedArgs = args
+if (typeof parsedArgs === 'string') {
+  try { parsedArgs = JSON.parse(parsedArgs) } catch { parsedArgs = {} }
+}
+const overrides = (parsedArgs && typeof parsedArgs === 'object' && !Array.isArray(parsedArgs)) ? parsedArgs : {}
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
APPROVED

Normalizes `args` from a JSON-encoded string back to an object before merging over `DEFAULT_UNIT`, so custom units passed via `/studio-implement` actually drive the loop instead of silently falling back to defaults.

The normalization is correct and backward-compatible: an object arg still works (skips the parse), `undefined` and non-JSON strings fall through to `{}` and resolve to `DEFAULT_UNIT`, and a JSON array is rejected by the existing `!Array.isArray` guard. `args` is consumed nowhere else in the file, so this is the only call site that needed it. The `Unit:` log line gives a clean end-to-end confirmation on the next run.

**Must fix**
None.

**Consider**
- Nothing blocking. Posted as a comment rather than an approval because the PR is self-authored and GitHub blocks self-approval.

eladrianovalle (commented):
APPROVED

Correctly normalizes `args` when the runtime forwards it as a JSON-encoded string, so custom units actually override `DEFAULT_UNIT` instead of silently falling back.

**Must fix**
- None.

**Consider**
- The `typeof parsedArgs === 'object' && !Array.isArray(parsedArgs)` guard already catches the case where the string parses to a non-object (number, array, null), so partial/garbage args degrade safely to defaults — good. No change needed; just noting the behavior is intentional.

eladrianovalle:
Superseded by #33 — the args fix is folded into the single 'wire config into the loop' PR (it was discovered while doing that work). Closing to keep one coherent change.

Issue #3, opened 2026-03-14
Fix 3 theater tests in test_validators.py


eladrianovalle:
## Summary

Three `validate_run` tests assert only `isinstance(result, ValidationResult)` — they would pass even if the entire validation logic were deleted. These are green checkmark theater.

**Source:** Test Engineer advocate/contrarian R2, Integrator Item 12 — from `run_studio_20260314_034202`

## Tests to fix

1. `test_validate_run_complete` (~line 351): Replace `assert isinstance(result, ValidationResult)` with assertions on `result.passed`, `result.issues`, and warning content
2. `test_validate_run_studio_phase` (~line 521): Same — assert meaningful properties
3. `test_validate_run_tech_phase` (~line 534): Same

## Mutation verification

Replace the entire body of `validate_run` with `return ValidationResult(passed=True, issues=[])`. All three tests currently still pass. After fix, they must fail.

## Effort estimate
~1 hour

🤖 Generated with [Claude Code](https://claude.com/claude-code) from studio run `run_studio_20260314_034202`

Issue #2, opened 2026-03-14
Wire DocumentValidator into _validate_artifacts() during finalize


eladrianovalle:
## Summary

The single highest-impact improvement from the Studio self-evaluation run. `_validate_artifacts()` in `run_phase.py` currently checks file *presence* only. `DocumentValidator` exists in `validators/document_validator.py` and can check completeness, consistency, format, and verdict presence — but it is not called during finalize.

**Source:** Design advocate R2 (Proposal 3b), Art advocate R2, Integrator Item 6 — from `run_studio_20260314_034202`

## What to do

- In `_validate_artifacts()`, after confirming files exist, call `DocumentValidator` checks as **warnings** (not hard failures):
  - `extract_verdict()` cross-check: warn if contrarian file lacks VERDICT string
  - Minimum length warning: flag advocate/contrarian files under ~200 characters as suspiciously short (rubber-stamp detector)
  - `check_format()`: verify markdown structure basics
- Add a quality score block to `run.json` with validation results
- Keep these as warnings, not hard failures — don't break existing workflows

## Acceptance criteria

- [ ] `finalize` calls DocumentValidator on advocate/contrarian files
- [ ] Warnings printed to stdout for missing verdicts, short files, format issues
- [ ] `run.json` includes a `quality` block with pass/warn/fail counts
- [ ] Existing tests still pass
- [ ] New tests verify the validation-during-finalize path

## Effort estimate
~2 days

🤖 Generated with [Claude Code](https://claude.com/claude-code) from studio run `run_studio_20260314_034202`

PR #33, opened 2026-06-28
feat: wire loop config into the implementation loop (bridge + consumption + args fix)

eladrianovalle:
## What

Closes the unused-config gap from the loop's Phase 2: `impl_loop.py`'s config is now **consumed end-to-end**, not just loadable. Includes the args-plumbing fix discovered while doing it (**supersedes #32**).

## The three layers

1. **Bridge** (`studio/impl_loop.py`) — built *by the loop itself* (`/studio-implement`): `runtime_knobs(config)` projects a `LoopConfig` into the workflow's knobs, and `python -m impl_loop` prints them as JSON. Writer + editor commits; editor cut 1 redundant test (19→18, green).
2. **Consumption** (`.claude/workflows/implementation-loop.js`) — the workflow honors the knobs: `editor.mandate="off"` → skips the editor pass (delivers writer's version); `read_scope`/`output_budget` injected into the editor prompt.
3. **Wiring** (`.claude/commands/studio-implement.md`) — `/studio-implement` reads `python -m impl_loop`, merges `editor_enabled`/`read_scope`/`output_budget` into the unit args, and invokes the loop by **`scriptPath`**.

## Args fix (was #32)

The loop ignored custom units across three runs. Two compounding harness gotchas, both pinned with ~$0 no-agent probes:

- **`Workflow({name})` resolves a FROZEN registry snapshot** — it never saw on-disk edits to the workflow. Fix: invoke by `scriptPath` (live file).
- **`args` arrives as a JSON string**, not an object. Fix: the workflow `JSON.parse`s it.

Both are now documented in the command's "Gotchas" section so they don't regress.

## Tests

`cd studio && python -m pytest tests/test_impl_loop.py -q` → 18 passed. `python -m impl_loop` emits valid knobs JSON. Workflow + probe verified that custom args resolve via `scriptPath`.

### DIFF
--- a/.claude/commands/studio-implement.md
+++ b/.claude/commands/studio-implement.md
@@ -39,6 +39,7 @@ From `$ARGUMENTS`, construct the workflow `args`:
   - Node (`package.json`): `npm test` or the project's test script.
   - Rust (`Cargo.toml`): `cargo test`.
   - Unity / other: ask if you cannot infer a runnable command.
+  - If you cannot infer one, fall back to the config knob `test_command` from `python -m impl_loop` (Step 4).
 - `static_check` — `ruff check <path>` for Python; the project linter otherwise; omit if none.
 
 ### Step 2 — Pick the target branch (safety)
@@ -82,14 +83,21 @@ python -m impl_loop   # prints knobs JSON, e.g. {"editor_enabled": true, "read_s
 ```
 
 Then call the **Workflow** tool **by `scriptPath`** (NOT `name` — see Gotchas), merging the
-relevant knobs (`editor_enabled`, `read_scope`, `output_budget`) into the unit args:
+config knobs into the unit args:
 
 ```
 Workflow({ scriptPath: "<repo>/.claude/workflows/implementation-loop.js", args: {
-  unit_id, title, instructions, test_command, static_check,
-  editor_enabled, read_scope, output_budget   // from `python -m impl_loop`
+  unit_id, title, instructions, static_check,
+  test_command,            // per-unit inferred; if none, use the knob test_command
+  editor_enabled, read_scope, output_budget,           // from `python -m impl_loop`
+  static_checks, require_mutation_check                // from `python -m impl_loop`
 }})
 ```
+
+How the workflow consumes each knob: `editor_enabled=false` → skip the editor pass;
+`read_scope`/`output_budget` → shape the editor prompt; `require_mutation_check=false` → writer
+skips the mutation check; `static_checks=[]` → writer skips the static check and the entry gate
+doesn't require it. So passing all knobs makes `implementation_loop.toml` fully live.
 
 The workflow runs in the background and notifies on completion. Do not re-run it or poll;
 wait for the completion notification.
--- a/.claude/commands/studio-implement.md
+++ b/.claude/commands/studio-implement.md
@@ -72,19 +72,37 @@ Print a short block, then proceed (unless `--plan`):
 
 If `--plan`, stop here.
 
-### Step 4 — Run the loop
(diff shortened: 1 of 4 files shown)


eladrianovalle (commented):
APPROVED

Wires the loop config through end-to-end (knobs bridge → workflow consumption → command wiring) plus the `scriptPath`/JSON-string args fix; clean and well-covered by tests.

**Must fix**
- None.

**Consider**
- `impl_loop.py:runtime_knobs` emits `test_command`, `static_checks`, and `require_mutation_check`, but `studio-implement.md` Step 4 only merges `editor_enabled`/`read_scope`/`output_budget` into the workflow args. The other three are printed and never consumed — harmless, but a reader will expect them to do something. Either plumb them or drop them from the dict.
- `implementation-loop.js:~148` uses `u.output_budget || 400`, so an explicit `0` falls back to 400. A zero word-budget is nonsensical anyway, so this is fine — just flagging the falsy-coalesce.
- The args-string normalization and `editor_enabled === false` short-circuit both default safely (object passthrough, and editor runs unless explicitly disabled). Good defensive choices.

eladrianovalle (commented):
APPROVED

Closes the unused-config gap cleanly: `runtime_knobs` projects the already-resolved `LoopConfig` into the workflow knobs, the workflow honors them (editor skip, read-scope, output-budget), and the `scriptPath`/`JSON.parse` args fixes are correct and well-documented. Tests cover both default and override paths.

**Must fix**
- None.

**Consider**
- `output_budget || 400` treats a configured `0` as "unset" (falls back to 400). Almost certainly irrelevant given config validation, just noting the truthiness coupling.
- The editor-prompt branch keys off `read_scope === 'touched'`, so any non-`touched` value implies importers. That's exactly right today because the config validates `read_scope` against `VALID_READ_SCOPES`; if a third scope is ever added, revisit this branch.

eladrianovalle:
Addressed review item #1 (unconsumed knobs) in `323600c` — chose to **plumb** rather than drop, since spec §4 intends `test_command` as the config-level gate knob:

- `require_mutation_check=false` → writer skips the mutation check.
- `static_checks=[]` → writer skips the static check and the entry gate stops requiring `static_ok`.
- `test_command` → `/studio-implement` falls back to the config knob when it can't infer a per-unit command.
- `/studio-implement` now merges all knobs into the args and documents how each is consumed.

No `impl_loop.py` change — `runtime_knobs` already emitted these; this is the consume side. Items #2 (`output_budget || 400` falsy-coalesce) and #3 (`read_scope` binary branch) left as-is per the review (cosmetic / guarded by `VALID_READ_SCOPES`).

eladrianovalle (commented):
APPROVED

Follow-up push `323600c` plumbs the three previously-unconsumed knobs through the loop, closing the only Consider item from my last review.

**Must fix**
- None.

**Consider**
- The branching reads cleanly and the defaults are safe: undefined `require_mutation_check` keeps the mutation check on, and `static_checks` only relaxes the writer prompt + entry gate when it's an explicit `[]`. Good.
- `staticRequired` keys off `static_checks.length === 0`; a non-empty `static_checks` still drives the prompt off the singular `static_check` string. Fine today since the two travel together, just worth keeping aligned if the schema ever consolidates on the array.

eladrianovalle (commented):
APPROVED

Follow-up push `bfc0a1a` adds a clarifying comment on the `static_checks`/`static_check` duality — the exact alignment risk I flagged last review. Doc-only, no logic change; the entry-gate behavior is identical.

**Must fix**
- None.

**Consider**
- Nothing further. The note captures the intent well and the two stay aligned for the current schema.

PR #36, opened 2026-06-28
feat: add web_product role to manifest

### LINKED ISSUE #35: web_product role missing from studio.manifest.json — clobbers downstream packs on update
The manifest defines `web_engineering`, `web_test_engineer`, and `web_qa`, but **not `web_product`** (Director of Product — Web Platforms & Creator Tools), even though it's part of the same web-role set used by downstream `orcpunk_web`-style packs.

**Impact:** `web_product` only exists in downstream installs' `studio.manifest.json`. Since `studio.manifest.json` is in `install.py`'s `SOURCE_FILES`, every `update`/`init` overwrites the downstream manifest with upstream's — silently **dropping `web_product`** and breaking any pack that lists it (the role becomes undefined). Downstream then has to re-merge the block by hand after each update.

**Fix:** add the `web_product` role definition to `studio/studio.manifest.json` (alongside the other three `web_*` roles) so it's tracked upstream and survives updates.

PR incoming.

eladrianovalle:
Adds the **`web_product`** role (Director of Product — Web Platforms & Creator Tools) to `studio/studio.manifest.json`, alongside the existing `web_engineering` / `web_test_engineer` / `web_qa`.

**Why:** the role was only defined in downstream installs. Because `studio.manifest.json` is in `install.py`'s `SOURCE_FILES`, every `update`/`init` overwrote the downstream manifest with upstream's and silently dropped `web_product`, breaking any pack that lists it (e.g. an `orcpunk_web` pod) and forcing a manual re-merge after each update.

**Change:** one role definition added to the manifest (18 lines) + a CHANGELOG entry. No code changes.

Closes #35.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to TheGameStudio are documented here.
 - `/unstale` Agent 5 — project tracking audit (GitHub Issues + local tracking files) with permission-gated destructive actions
 - Stack-agnostic `/unstale` — the command now self-detects the project stack (Rust/Unity/Node/Python/Go) from marker files instead of assuming the Studio Python layout, so it works in any installed repo. Optional per-repo `.studio/unstale.toml` override pins exact snapshot commands and audit globs; authored by a new setup wizard `unstale_config` step (`CURRENT_SETUP_VERSION` 3) via `suggest_unstale_from_stack`
 - `ai_engineer` role in manifest — Staff AI Engineer specializing in prompt architecture & agent optimization
+- `web_product` role in manifest — Director of Product (Web Platforms & Creator Tools): opportunity framing (ICP/JTBD), build-vs-fork-vs-buy with licensing implications, MVI-gated roadmaps. Completes the `web_*` role set so downstream packs that list it survive updates (#35)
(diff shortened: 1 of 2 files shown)


eladrianovalle (commented):
APPROVED

Adds the `web_product` role to `studio/studio.manifest.json` so it survives `install.py` updates, completing the `web_*` set — clean, data-only change with a matching CHANGELOG entry.

**Must fix**
- None.

**Consider**
- Nothing blocking. The new entry's schema matches the sibling `web_*` roles exactly (same six keys, empty `prompt_doc` like the other 13), JSON parses, and CI (`triage`, no-conflicts) is green. Verified the root cause holds: `studio.manifest.json` is in `SOURCE_FILES`, so the upstream copy is what gets reinstalled — defining the role here is the right fix for #35.

eladrianovalle (commented):
APPROVED

Adds the `web_product` role to the manifest with a schema identical to the other `web_*` roles; valid JSON, both checks green, no code paths touched.

**Must fix**
- None.

**Consider**
- `prompt_doc` is `""`, consistent with the sibling `web_*` roles. If/when these roles get standalone prompt docs, this one will want one too — fine to defer.

PR #37, opened 2026-06-29
feat(install): ship the implementation loop to all installed repos

eladrianovalle:
## Why

`/studio-implement` worked only in the Studio source repo — the cross-repo installer shipped none of its machinery (no command, no `impl_loop.py`/config, no workflow-copy mechanism at all). Now that the loop is proven, open it to every repo that installs Studio.

## Changes

**Installer (`install.py`)** — `init` and `update` (which delegates to `install_studio`):
- `SOURCE_FILES` += `impl_loop.py` + `config/implementation_loop.toml` → `.studio/source/`.
- `SLASH_COMMANDS` += `studio-implement.md`.
- New `WORKFLOW_FILES` + a copy step → `.claude/workflows/implementation-loop.js` (verbatim, mirroring how commands are shipped).

**Path portability** — these files are copied verbatim, so they must resolve the *installed* layout (`.studio/source/`, not `studio/`):
- `studio-implement.md` runs `.studio/source/impl_loop.py` (falls back to `studio/impl_loop.py` in the source repo) and points `scriptPath` at the repo's own `.claude/workflows/implementation-loop.js`.
- `implementation-loop.js` editor prompt reads `CONTRARIAN_MANDATE` from `.studio/source/scopes.py` (or `studio/scopes.py`).

**Target-repo config override** — `impl_loop.py`'s CLI (`_cli`) now takes an optional config path, so the command passes `<repo>/.studio/implementation_loop.toml` when present (the project override lives at the target root, outside the snapshot's resolution chain). Loud on a typo'd path; defaults otherwise.

## Tests

- Install ships source + config + command + workflow (verbatim match).
- CLI: explicit path flows through / loud on missing / defaults with no arg.
- **607 pass** (full suite).

## Rollout

Existing installs pick it up via `/studio-update` (the new source files show as an available update; the command + workflow copy on re-install).

### DIFF
--- a/.claude/commands/studio-implement.md
+++ b/.claude/commands/studio-implement.md
@@ -75,22 +75,27 @@ If `--plan`, stop here.
 
 ### Step 4 — Load config knobs, then run the loop
 
-First read the loop config so the run honors `implementation_loop.toml` (run from the `studio/`
-package dir):
+**Studio path:** use `.studio/source/impl_loop.py` for the command below. If that file does not
+exist but `studio/impl_loop.py` does, use that instead (you are in the Studio source repo).
+
+Read the loop config knobs. If the repo has a project override at `.studio/implementation_loop.toml`,
+pass it explicitly (it lives at the repo root, outside the snapshot's resolution chain); otherwise
+run with no arg (shipped default → built-in defaults):
 
 ```bash
-python -m impl_loop   # prints knobs JSON, e.g. {"editor_enabled": true, "read_scope": "touched+importers", "output_budget": 400, ...}
+python .studio/source/impl_loop.py [.studio/implementation_loop.toml]
+# prints knobs JSON, e.g. {"editor_enabled": true, "read_scope": "touched+importers", "output_budget": 400, ...}
 ```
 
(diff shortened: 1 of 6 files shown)


eladrianovalle (commented):
APPROVED

Ships the writer/editor implementation loop to every Studio-installed repo (source + config + command + workflow), with verbatim copies that resolve the installed `.studio/source/` layout at run time.

**Must fix**
- None.

**Consider**
- The new `WORKFLOW_FILES` copy mirrors the slash-command copy precisely (`studio_dir.parent/.claude/workflows`, same `samefile` self-copy guard) — good. One trivial inconsistency: commands `mkdir` the dest once before the loop while workflows `mkdir` inside the loop per-file. Both are correct; the workflow form is actually nicer (no empty dir when the list is empty), so no change needed unless you want them to read identically.
- Like the slash commands, copied workflows aren't tracked in `MANIFEST.json` checksums, so `/studio-update` won't surface drift in `implementation-loop.js` the way it does for source files — it just re-copies. That matches the established command behavior, so fine to leave; worth a mental note if workflows ever need update-detection.

PR #38, opened 2026-06-29
fix: cross-repo CLI hardening (config resolution, error UX, hints, install gaps)

eladrianovalle:
## Why

Using Studio from *installed* repos kept producing little friction points — CLI quirks that made agents bulldoze past the intended process instead of following it. A fan-out review (5 reviewers + adversarial verification) surfaced **21 confirmed findings**, nearly all the same shape: *works in the source repo, silently breaks or confuses in an installed repo*. This fixes the whole cluster.

## The headline bug — project config silently ignored

`_resolve_scopes` (and `validation.toml`) read project-local config relative to `get_studio_root()`, which in an installed repo is `<repo>/.studio/source/` — so they probed `…/.studio/source/.studio/scopes.toml` (never exists) and fell back to shipped defaults. But the setup wizard writes to `<repo>/.studio/`. **`/studio-setup`-tuned scope budgets were silently dropped on every installed run.** Now both read from the artifact root, like roles/personas/clarity/integrations already do.

## `get_artifact_root` rewrite (H2 / H7 / M2 / L1)

- Recognizes the installed snapshot: `<repo>/.studio/source` → repo root (never writes into the snapshot).
- Walks up for `.studio/VERSION`, so a call from a monorepo subdir resolves to the repo root instead of scaffolding a second unconfigured `.studio/`.
- Only warns on a genuine fresh external run (no more spurious "defaulting to cwd" on every correct install).

## The "bulldoze" root cause — error/hint UX

- `main()` now surfaces operational failures as `Error: <msg>` + exit 1 instead of a raw traceback (`STUDIO_DEBUG=1` restores it).
- `tomli`-missing on 3.10 → actionable `pip install tomli` message.
- `finalize`'s next-step `validate` hint was broken (missing `--phase`, unresolvable path); all hints now use `_entrypoint()` so they copy-paste in source **and** installed repos.
- `inject-context` warns instead of silently dropping scope guidance when a recorded path doesn't resolve.

## Slash commands / install

- `run-studio-phase.md`: `prompt_doc` path is repo-root-relative (the stale `.studio/source/` prefix from the role-prompt decoupling skipped installed personas).
- `studio-update.md`: detect the snapshot `WARNING` and route the user to update from upstream.
- Installer ships `docs/SCOPES_GUIDE.md` (the CLI pointed at it but never shipped it); `init` announces a created `CLAUDE.md`; the bridge doc is created even for `init`'d repos (the `.studio`-exists guard had skipped it).

## Tests

Artifact-root snapshot/subdir resolution, scopes-from-artifact-root, bridge-on-init, SCOPES_GUIDE shipped. **612 pass** (full suite).

*Review cost: 32 agents, ~1M tokens; 6 findings rejected by the verifier as not strictly cross-repo (a few are real CLI-ergonomic quirks — e.g. no machine-readable `prepare`/`extract-decisions` output — left for a separate, more invasive pass).*

### DIFF
--- a/.claude/commands/run-studio-phase.md
+++ b/.claude/commands/run-studio-phase.md
@@ -152,7 +152,7 @@ python ".studio/source/run_phase.py" inject-context --run-dir {run_dir} --scope 
 > - {List specific S2 files this role needs per the dependency map}
 > - Condensed brief of all other roles: `{run_dir}/S2-brief.md`
 >
-> {If this role has a prompt doc, read it at `.studio/source/{prompt_doc}` for detailed guidance.}
+> {If this role has a prompt doc (the `prompt_doc` shown in the Role Menu), read it at that path — it is relative to the repo root (a doc that lives in THIS project), not the Studio snapshot.}
 >
 > Write a thorough proposal covering all required deliverables. No word cap — this is the full analysis.
 > Save to `{run_dir}/advocate--{role}--S2-01.md`.
--- a/.claude/commands/studio-update.md
+++ b/.claude/commands/studio-update.md
@@ -20,6 +20,12 @@ python ".studio/source/run_phase.py" check-install --target .
 
 If already up to date, tell the user and stop.
 
+**If the output contains a `WARNING:` line** (e.g. "running from the installed snapshot … cannot
+compare against live source"), the check/update is running against the snapshot and cannot see the
+real upstream — so it may report "up to date" when it isn't. In that case, tell the user to run the
+update from the upstream Studio source repo instead:
+`python studio/run_phase.py update --target <this-repo>`. Do not trust the result until then.
+
 ### Step 3: Update
 
 ```bash
--- a/studio/impl_loop.py
+++ b/studio/impl_loop.py
@@ -13,7 +13,13 @@ from __future__ import annotations
 try:
     import tomllib  # Python 3.11+
 except ModuleNotFoundError:
-    import tomli as tomllib  # type: ignore[no-redefine]  # Python 3.10 fallback
+    try:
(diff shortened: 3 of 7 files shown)


eladrianovalle (commented):
APPROVED

Solid, surgical cross-repo hardening — the installed-vs-source path resolution is correct and the diff is well-covered by new tests.

Verified the core risk (a regression in the source repo): in the source tree `get_artifact_root()` still equals `get_studio_root()` (since `_installed_repo_root` only fires when `studio_root` is `<repo>/.studio/source`), so the `_resolve_scopes` / `validation.toml` switch to `get_artifact_root()` is a no-op there and only redirects config reads to `<repo>/.studio/` in installed repos — which is exactly where the setup wizard writes. The snapshot-detection, walk-up-to-VERSION, and bridge-doc-on-init paths all have direct tests.

**Must fix**
- None.

**Consider**
- `main()` now collapses every `Exception` from `_dispatch` into `Error: <msg>` + exit 1. `SystemExit` (incl. the new tomli `raise SystemExit(...)`) is `BaseException`, so it correctly passes through — good. Just confirm no caller/test relied on a specific exception type or a non-1 exit code propagating out of `main()` (612 green suggests not).
- `finalize_run`'s new hint passes `--phase {phase}`; fine as long as `validate` actually accepts `--phase` (the old hint omitting it was the bug being fixed). Worth a one-line smoke of `validate --phase ... --run-id ...` before relying on the copy-paste.
- `_find_installed_root_upwards` walks to the first ancestor with `.studio/VERSION`; in a nested-install edge (an inner init'd repo under an outer one) it resolves to the nearest, which is the right call — just noting the precedence is intentional.

PR #39, opened 2026-06-29
feat(cli): machine-readable output + derivable --phase (ergonomic pass)

eladrianovalle:
## Why

The cross-repo review (#38) verifier rejected 6 findings as "real but not cross-repo-specific." Three of those are genuine CLI-ergonomic quirks that make agents work *around* the process — they just bite equally everywhere, not only in installed repos. This is that cluster (R3/R4/R6). Stacked on #38.

## Changes

- **`prepare --json`** (R6) — emits `{run_id, run_dir, instructions, phase, scoped, objective_changed}` as the **final line of stdout** (cleanup/scaffold notices precede it; the storage tip is suppressed under `--json`). Agents parse it instead of scraping the prose summary that every downstream `--run-dir` step depended on.
- **`extract-decisions --json`** (R3) — emits `{count, decisions:[...]}` so the pause/proceed gate keys off `count`, not "is stdout empty." Mirrors the existing `check-decisions`/`stats --json`.
- **`--phase` now optional** on `finalize` / `validate` / `recompute-clarity` (R4) — derived from `run_id` (`run_<phase>_<ts>`) via `_phase_from_run_id`; a malformed `run_id` errors clearly. Removes the redundant-arg friction (phase was always encoded in `run_id`).

## Commands wired to use it

`run-phase.md` and `run-studio-phase.md` now call `prepare --json` (parse the last line) and gate decision pauses on `count > 0` — no more prose-scraping or empty-stdout detection.

## Skipped from the 6 (per the earlier analysis)

R1 (`studio-setup` hardcodes `.studio/source/` — only breaks the *source* repo), R2 (`<repo>` placeholder — layout-neutral doc gap, friction claim was unsupported), R5 (rating `input()` — already TTY-guarded + both commands pass `--no-rate-prompt`).

## Tests

`_phase_from_run_id` (derive + raise), `prepare --json` shape, `finalize` without `--phase`, `extract-decisions --json` count. **616 pass.**

### DIFF
--- a/.claude/commands/run-phase.md
+++ b/.claude/commands/run-phase.md
@@ -18,12 +18,14 @@ Run the prepare command to create a run directory and instructions:
 **Studio path:** Use `.studio/source/run_phase.py` for all commands below. If that file does not exist but `studio/run_phase.py` does, use `studio/run_phase.py` instead (you are in the Studio source repo).
 
 ```bash
-python ".studio/source/run_phase.py" prepare $ARGUMENTS
+python ".studio/source/run_phase.py" prepare $ARGUMENTS --json
 ```
 
 If running from a repo that is NOT the Studio repo itself, artifacts will automatically land in the current repo under `.studio/output/`. A bridge doc will be created on first use.
 
-Note the run_id and run directory path from the output.
+`--json` prints a machine-readable object as the **final line of stdout** — parse it for
+`run_id`, `run_dir`, and `instructions` rather than scraping prose:
+`{"run_id": "...", "run_dir": "...", "instructions": "...", "phase": "...", ...}`.
 
 ### Step 2: Read Instructions
 
@@ -71,10 +73,12 @@ python ".studio/source/run_phase.py" record-metrics --run-dir {run_dir} --agent 
 
 Then extract decision points:
 ```bash
-python ".studio/source/run_phase.py" extract-decisions --run-dir {run_dir}
+python ".studio/source/run_phase.py" extract-decisions --run-dir {run_dir} --json
 ```
 
-**If the output is non-empty (any decision points found), you MUST pause and present them ALL to the user before proceeding.** Do not continue to the contrarian until the user has responded.
+`--json` emits `{"count": N, "decisions": [...]}`. **If `count > 0`, you MUST pause and present
+every decision to the user before proceeding.** Do not continue to the contrarian until the user
+has responded. (Gate on `count`, not on whether stdout is empty.)
 
 Present each decision to the user:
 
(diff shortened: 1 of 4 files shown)


eladrianovalle (commented):
APPROVED

A tight, well-tested ergonomic pass: `--json` machine output on `prepare`/`extract-decisions` and a derivable `--phase` on the run-scoped commands, with the prose paths preserved.

Verified the core risks:
- `prepare --json` stays the **last** stdout line: the only later prints are the usage-log append (no stdout) and the storage tip, which is now correctly gated by `and not emit_json`. Scaffold/cleanup notices precede the object, so "parse the final line" holds.
- `extract-decisions --json` returns before the silent-empty branch, so `{"count": 0, "decisions": []}` is always emitted — the doc's "gate on `count`, not empty stdout" is honored.
- `_phase_from_run_id` tolerates the underscore-bearing timestamp (`run_<phase>_<ts>` → `parts[1]`) and raises a clear `ValueError` on a malformed id; the `args.phase or _phase_from_run_id(...)` fallback keeps an explicit `--phase` authoritative (argparse `choices` still validates it).

**Must fix**
- None.

**Consider**
- `run-phase.md` "Brief update cadence" still calls `extract-decisions --run-dir {run_dir}` without `--json` (the gather-all step). Harmless, but it's now the lone prose-scraping holdout in a doc that otherwise keys on `count`.
- `prepare --json` is single-line (compact) while `extract-decisions --json` uses `indent=2` (multi-line). Both parse fine and extract is the sole output of its command, but a consumer wiring both might expect uniform formatting.

eladrianovalle (commented):
APPROVED

Re-review of the new push (`9890b40`, "simplify-pass cleanups") on top of the already-approved ergonomic work — a pure refactor, no behavior change, suite still green (616 pass).

The two cleanups both check out:
- `get_artifact_root()` — `installed_root = _installed_repo_root(studio_root)` moved inside the `cwd within studio_root` branch. That's its only use site (lines 304–305), so the move just skips the call on the other paths; the added comment correctly documents why the VERSION-walk-up branch and the bare-`.studio/` cwd branch stay separate.
- `_decision_to_dict(dp)` — shared serializer now backs both `check-decisions` and `extract-decisions --json`. Output shapes are preserved: `extract-decisions` keeps the same field order, and `check-decisions` does `grouped[entry.pop("priority")]` so `priority` is dropped from the per-entry dict exactly as before. The popped dict is freshly built per call, so no aliasing.

**Must fix**
- None.

**Consider**
- The two earlier "Consider" notes (the `extract-decisions` prose-scraping holdout in `run-phase.md`, and compact-vs-indented JSON formatting) are untouched by this push — still non-blocking, just flagging they carry forward.

eladrianovalle (commented):
APPROVED

Re-review of the new push (`2c7ddd7b`): it merges `origin/main` into the branch and carries **no file changes** — the compare against the last-reviewed head (`9890b40`) is empty, since this PR is stacked on #38 and main's content was already present. Pure branch-sync, no behavior change. The ergonomic work approved on `ef0d9ee`/`9890b40` stands.

**Must fix**
- None.

**Consider**
- The two prior non-blocking notes carry forward unchanged: the `extract-decisions --run-dir` prose-scraping holdout in `run-phase.md`, and compact-vs-`indent=2` JSON formatting between `prepare` and `extract-decisions`.

PR #40, opened 2026-06-29
fix(impl_loop): resolve project config override against the artifact root

eladrianovalle:
## Why

The `/simplify` altitude review caught that `impl_loop.py` carries the **same config-resolution bug** the scopes/validation fix (#38) removed — one module deeper. `_resolve_config_path` resolved *both* the project override (`.studio/implementation_loop.toml`) and the shipped default against `STUDIO_ROOT`, which is `<repo>/.studio/source` in an installed repo. So the override was sought at `<repo>/.studio/source/.studio/implementation_loop.toml` — never where the wizard writes it (`<repo>/.studio/`). It was papered over by `/studio-implement` passing the path explicitly, meaning any other caller silently got shipped defaults.

## Fix

- New `_project_artifact_root(studio_root)` maps an installed snapshot (`<repo>/.studio/source` → `<repo>`) and honors `STUDIO_ARTIFACT_ROOT` — mirroring `run_phase.get_artifact_root`'s detection **without importing run_phase** (`impl_loop` ships standalone to `.studio/source/`).
- The project override now resolves at the consuming repo root; the shipped default stays under the source dir.
- Removed the `/studio-implement` explicit-path crutch (resolution handles it). The CLI's optional path arg stays for non-standard locations.

## Verified

- New tests: installed-layout override resolves at the repo root (the regression), `STUDIO_ARTIFACT_ROOT` override. **618 pass.**
- End-to-end smoke: a real `.studio/source/impl_loop.py` with an override at the repo root and **no path arg** now emits the override (`editor_enabled:false, output_budget:77`) instead of shipped defaults.

Stacked on #39.

### DIFF
--- a/.claude/commands/studio-implement.md
+++ b/.claude/commands/studio-implement.md
@@ -78,12 +78,12 @@ If `--plan`, stop here.
 **Studio path:** use `.studio/source/impl_loop.py` for the command below. If that file does not
 exist but `studio/impl_loop.py` does, use that instead (you are in the Studio source repo).
 
-Read the loop config knobs. If the repo has a project override at `.studio/implementation_loop.toml`,
-pass it explicitly (it lives at the repo root, outside the snapshot's resolution chain); otherwise
-run with no arg (shipped default → built-in defaults):
+Read the loop config knobs. Resolution finds a project override at the repo root
+(`.studio/implementation_loop.toml`) automatically, falling back to the shipped default →
+built-in defaults — so no path argument is needed:
 
 ```bash
-python .studio/source/impl_loop.py [.studio/implementation_loop.toml]
+python .studio/source/impl_loop.py
 # prints knobs JSON, e.g. {"editor_enabled": true, "read_scope": "touched+importers", "output_budget": 400, ...}
 ```
 
--- a/studio/impl_loop.py
+++ b/studio/impl_loop.py
@@ -21,6 +21,7 @@ except ModuleNotFoundError:
             "Install it with: python -m pip install tomli  (or upgrade to Python 3.11+)."
         )
 import json
+import os
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import List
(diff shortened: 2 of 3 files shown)


eladrianovalle (commented):
APPROVED

Carries the #38 installed-layout config-resolution fix one module deeper into `impl_loop.py`, so the project override now resolves at the consuming repo root even when the module runs from a `.studio/source` snapshot.

Verified the core risk (no regression in the source repo): `_project_artifact_root` only remaps when `studio_root` is `<repo>/.studio/source` (`name == "source"` and parent `.studio`); otherwise it returns `studio_root` unchanged, so in the source tree `local` stays `studio_root/.studio/implementation_loop.toml` exactly as before. The installed branch redirects to `<repo>/.studio/`, where the wizard writes — matching `run_phase.get_artifact_root`. Both new tests (installed override at repo root; `STUDIO_ARTIFACT_ROOT`) and the doc/`--studio-implement` crutch removal line up with the resolution change.

**Must fix**
- None.

**Consider**
- `_project_artifact_root` deliberately re-implements `run_phase.get_artifact_root`'s detection (documented: impl_loop ships standalone) — the two can now drift independently if one's detection rule changes. Worth a comment cross-link or a shared test asserting parity.
- Minor asymmetry: the `STUDIO_ARTIFACT_ROOT` branch does `.expanduser().resolve()` while the snapshot branch returns `studio_root.parent.parent` unresolved. Harmless, just inconsistent.

eladrianovalle (commented):
APPROVED

Carries the #38 config-resolution fix one module deeper: `impl_loop`'s project override now resolves at the consuming repo root instead of under the installed `.studio/source` snapshot, so any caller (not just `/studio-implement` passing the path explicitly) honors it.

Verified the core logic:
- `STUDIO_ROOT = Path(__file__).resolve().parent` is `<repo>/.studio/source` in an installed layout, so `_project_artifact_root` hits the `name == "source" and parent.name == ".studio"` branch and returns `<repo>` → override at `<repo>/.studio/implementation_loop.toml`, exactly where the setup wizard writes. That's the regression the new `test_..._installed_override_resolves_at_repo_root` pins.
- In the Studio source repo (`<repo>/studio/impl_loop.py`), `name == "studio"` so `_project_artifact_root` returns `studio_root` unchanged — same path the old code used, so no behavior change there. Shipped default still resolves via `studio_root / "config"`.
- `STUDIO_ARTIFACT_ROOT` short-circuits the lookup to an explicit root, covered by the second new test.
- The `/studio-implement` doc drops the explicit-path crutch in step with the resolution change, and the optional CLI arg is retained for non-standard locations.

**Must fix**
- None.

**Consider**
- `_project_artifact_root` is a deliberate partial mirror of `run_phase.get_artifact_root` — it has the env + installed-snapshot branches but not the walk-up-to-VERSION ancestor search. For the studio_root values this module actually sees (derived from `__file__`) the two agree, but if a future layout lets them diverge, the override could be sought in a different place than where the wizard wrote it. Worth a one-line note pointing the two implementations at each other so they're kept in sync.

eladrianovalle (commented):
APPROVED

Correct, surgical fix: `impl_loop.py` now resolves the project override (`.studio/implementation_loop.toml`) at the consuming repo root instead of under the source snapshot, so the `/studio-implement` explicit-path crutch is no longer needed — matching the #38 fix one module deeper.

Verified the core logic:
- In an installed layout `STUDIO_ROOT = <repo>/.studio/source` (`Path(__file__).resolve().parent`, line 30), so `_project_artifact_root`'s `name == "source" and parent.name == ".studio"` branch fires and maps to `<repo>` — the override lands at `<repo>/.studio/implementation_loop.toml`, where the wizard writes it. The shipped default stays at `<studio-root>/config/`, correctly under the snapshot.
- Source-repo fallback returns `studio_root` unchanged, so `get_artifact_root() == get_studio_root()` parity holds there and the pre-existing `studio_override_beats_shipped_default` test is unaffected (its tmp `studio_root` has neither the `source`/`.studio` names nor the env var → identity). No regression.
- `STUDIO_ARTIFACT_ROOT` takes precedence over both, scoping only the project-override lookup (not the shipped default) — same split as `run_phase`. Both new tests cover exactly these two paths (installed-layout regression + env override).

**Must fix**
- None.

**Consider**
- `_project_artifact_root` is now an independent second copy of `run_phase`'s installed-layout detection, minus the cwd-based `_find_installed_root_upwards` walk-up and the bare-`.studio/` cwd fallback. Harmless today because `impl_loop`'s `STUDIO_ROOT` is always the script's own `.studio/source` dir, so the snapshot mapping covers the real case — but the two can drift. The "don't import `run_phase`" constraint (standalone shipping) is reasonable; just noting the duplication is now load-bearing in two files.
- `impl_loop`'s module-level `STUDIO_ROOT` (line 30) ignores the `STUDIO_ROOT` env override that `run_phase.get_studio_root()` honors. Pre-existing and out of scope for this PR, just flagging the asymmetry since this change leans on `STUDIO_ROOT` for resolution.

eladrianovalle (commented):
APPROVED

Follow-up on the new push (`0eb40d64`): it's a clean sync-merge of `origin/main` into the branch — the working tree is byte-identical to the SHA I reviewed before (`71a4617`), and the net diff vs `main` is still the same three files (`impl_loop.py`, `studio-implement.md`, `test_impl_loop.py`, +69/−17). No new code, no conflicts, no Must-fix items from prior reviews to recheck. Approval carries over.

**Must fix**
- None.

**Consider**
- Unchanged from before: `_project_artifact_root` remains a deliberate standalone mirror of `run_phase.get_artifact_root`'s installed-layout detection — fine today, worth a cross-link comment so the two don't silently drift.

PR #41, opened 2026-06-29
feat(update): clobber-preview precondition + draft-PR etiquette

eladrianovalle:
Opened as a **draft** per the new Git/PR etiquette this PR adds. 🙂

## Why

You hit a case where the agent running `/studio-update` feared overwriting existing config. Validated: project config at `<repo>/.studio/*.toml` was always safe (update never writes there), but **edited files inside the snapshot** (`<repo>/.studio/source/**`, e.g. an edited shipped config) *do* get overwritten on update — because update re-installs the whole snapshot. This adds a real guard.

## Clobber-preview precondition

- **`check_studio` now returns `locally_modified`** — installed source files whose on-disk content has drifted from the checksum recorded at install (the clobber set). Distinct from `changed` (upstream-vs-installed = an update is available).
- **`update_studio` gains `force`** — when `locally_modified` is non-empty it returns `{blocked: True, locally_modified: [...]}` instead of overwriting.
- **`check-install`** surfaces a `⚠️ LOCAL EDITS` preview; **`update`** prints `Update BLOCKED` + the files + guidance (move project config to `<repo>/.studio/`, or re-run with `--force`).
- **`studio-update.md`** makes the preview a hard pre-condition and tells the agent **not** to `--force` on its own — surface to the human.

Tests: check flags an edited snapshot file; update blocks (edit survives); `--force` overrides. **620 pass.**

## Also: Git/PR etiquette in the injected principles

Added a section to `CODING_PRINCIPLES.md` (injected into every target repo's CLAUDE.md): **open PRs as drafts first** — prevents accidental early merges and reviewer churn on a moving target. Flip to ready when complete + green.

### DIFF
--- a/.claude/commands/studio-update.md
+++ b/.claude/commands/studio-update.md
@@ -26,13 +26,21 @@ real upstream — so it may report "up to date" when it isn't. In that case, tel
 update from the upstream Studio source repo instead:
 `python studio/run_phase.py update --target <this-repo>`. Do not trust the result until then.
 
+**If the output contains a `⚠️ … LOCAL EDITS` block** (installed snapshot files the user edited),
+this is the clobber preview — **STOP and surface it to the user before updating.** `update` will
+refuse to overwrite these unless `--force` is passed. For each listed file: if it's project config,
+the fix is to move it to `<repo>/.studio/<name>.toml` (the project-override location update never
+touches); otherwise confirm with the user whether to keep the edit (back it up) or discard it.
+
 ### Step 3: Update
 
 ```bash
 python ".studio/source/run_phase.py" update --target .
 ```
 
-Show the user what was updated (files changed, added, removed).
+If `update` prints `Update BLOCKED`, do NOT pass `--force` on your own — relay the listed files to
+the user and let them decide (move config out, back up, or explicitly approve `--force`). Otherwise
+show the user what was updated (files changed, added, removed).
 
 ### Step 4: Remind about restart
 
--- a/studio/docs/CODING_PRINCIPLES.md
+++ b/studio/docs/CODING_PRINCIPLES.md
@@ -60,6 +60,12 @@ For multi-step tasks, state a brief plan:
 
 Strong success criteria let you loop independently. Weak criteria ("make it work") require constant clarification.
 
+## 5. Git & PR Etiquette
+
+**Open pull requests as drafts first.**
+
+A new PR starts as a **draft**, not ready-for-review. Mark it ready only when the work is complete, tests pass, and you actually want a human to review and merge it. Opening as a draft prevents two failure modes: accidental early merges (a not-yet-finished PR getting merged by mistake) and reviewer churn (reviewers spending attention on a still-moving target). Flip to ready when it's genuinely ready for eyes.
+
 ---
 
 *These guidelines are working if:* fewer unnecessary changes in diffs, fewer rewrites due to overcomplication, and clarifying questions come before implementation rather than after mistakes.
--- a/studio/install.py
+++ b/studio/install.py
@@ -176,6 +176,25 @@ def _resolve_source_dir(
     return live, None
 
 
+def _recorded_upstream_source(version_path: Path, snapshot: Path) -> Optional[Path]:
+    """Return the upstream source dir recorded in an existing VERSION — but only if
+    it is still a usable upstream: it exists, has ``run_phase.py``, and is NOT the
+    target's own snapshot. Used to avoid overwriting a good pointer with a
+    self-pointing one when (re)installing from the snapshot.
+    """
+    try:
+        version = json.loads(version_path.read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError, ValueError):
+        return None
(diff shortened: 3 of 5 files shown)


eladrianovalle (commented):
APPROVED

Adds a real clobber guard: `check_studio` now reports `locally_modified` (snapshot files whose on-disk content drifted from the install-time checksum) and `update_studio` refuses to re-install over them unless `--force` is passed — cleanly surfaced in `check-install`, `update`, and the slash command.

The core logic holds up: `locally_modified` (on-disk vs recorded checksum) is correctly distinct from `changed` (upstream vs installed), the block check sits *after* the `up_to_date` early return so a no-op update never falsely blocks, every return path still carries `updated`/`added`/`removed`, and the all-or-nothing block is the right conservative call since `install_studio` re-installs the whole snapshot. Tests cover both the detection and the block→`--force` override paths.

**Must fix**
- None.

**Consider**
- The block is all-or-nothing: one unrelated locally-edited file blocks an otherwise-clean pending update. That's the safe default and the guidance (move config to `<repo>/.studio/`, or `--force`) covers it — just flagging it as intended behavior worth a line in the docs if users hit it often.
- `install.py:411-420` re-hashes every installed file for the `locally_modified` pass, separate from the earlier `changed`/`extra` walk. Fine at snapshot scale; only worth consolidating if the snapshot ever grows large.

eladrianovalle (commented):
APPROVED

Clobber-preview guard is correct: `update` now refuses to re-install over locally-edited snapshot files unless `--force`, with clear surfacing in `check-install`, `update`, and the `studio-update` command doc.

**Must fix**
- None.

**Consider**
- `check_studio` returns `locally_modified` as a separate dimension from `changed` (on-disk-vs-recorded vs upstream-vs-installed) — the docstring captures this well. A file that's both upstream-changed and locally-edited lands in both lists and still blocks, which is the safe outcome.
- When `up_to_date` is True, `update_studio` returns early and never reports `blocked` even if local edits exist. That's correct (nothing is overwritten, so no clobber), and `check-install` still surfaces the edits — just noting a caller relying solely on `update` output won't see the local-edits notice in that path.
- Draft is intentional per the etiquette this PR adds; nothing to action, leaving as discussion. Flip to ready when you want it merged.

eladrianovalle (commented):
APPROVED

Follow-up re-review of the single new commit (765ac49) since my last pass at 40b579e: a one-line `/simplify` cleanup in `_do_check_install`, dropping the defensive `status.get("locally_modified", [])` for a direct `status["locally_modified"]`.

Verified the contract this leans on: `check_studio` (install.py) returns `locally_modified` on *both* paths — the `installed: False` early return (install.py:382) and the normal return (install.py:432) — so the key is guaranteed present and the direct subscript can't KeyError. This also matches the sibling `_do_update` handler, which already subscripts directly, so the two CLI handlers now signal the same contract. No prior Must-fix items existed; nothing regressed.

**Must fix**
- None.

**Consider**
- None.

eladrianovalle (commented):
APPROVED

Follow-up re-review of the single new commit (b835ee8) since my last pass at 765ac49: a real bug fix preventing `install_studio` from recording a self-pointing `source_path` when (re)installing through a target's own snapshot.

The root cause is correctly identified and closed at the single write site: when `studio_dir.resolve() == source_dest.resolve()` (install/init run from inside the installed snapshot), the new `_recorded_upstream_source()` recovers a previously-recorded upstream pointer instead of clobbering it with the snapshot path — which would otherwise make every future snapshot-based check compare the snapshot against itself and silently report "up to date" with no breadcrumb upstream. The validity gate (exists + `run_phase.py` present + not the snapshot) mirrors `_resolve_source_dir`, the read is defensively wrapped so a corrupt/missing VERSION falls back to the snapshot pointer (same as the no-prior case), and the first-ever install-through-snapshot case still pins the snapshot — unavoidable, and covered by the #28 unresolvable-source warning. The `version_path` hoist is clean with no orphaned reassignment. The added regression test asserts the upstream pointer survives a snapshot re-install.

**Must fix**
- None.

**Consider**
- None.

eladrianovalle (commented):
APPROVED

Follow-up re-review of the single new commit (b835ee8) since my last pass at 765ac49: a guard so `install_studio` never records a self-pointing `source_path` when (re)installing FROM the snapshot.

Verified the fix: the new `_recorded_upstream_source` reads the *existing* VERSION (install.py:370) before it is overwritten (install.py:380), so there is no clobber-before-read. Its validity contract — `p.resolve() == snapshot or not (p / "run_phase.py").is_file()` (install.py:191) — exactly mirrors `_resolve_source_dir` (install.py:170), so a pointer this accepts is one that path will later trust. The `version_path` assignment was correctly hoisted above the read and the old duplicate removed. The degenerate case (install-from-snapshot with no prior/usable upstream pointer) still records the self-path, but that is unavoidable — no breadcrumb exists — and `_resolve_source_dir` already surfaces it as a loud warning rather than a silent "up to date". Test covers the preservation path. No prior Must-fix items; nothing regressed.

**Must fix**
- None.

**Consider**
- None.

PR #42, opened 2026-06-30
docs: unstale doc-sync — roles, counts, --force, dupes

eladrianovalle:
Output of a `/unstale` staleness audit (5 parallel agents: docs, code comments, memory, cross-refs, tracking). Code comments came back clean; these are the doc/tracking fixes.

## Fixed
- **`web_product` role** (the 14th manifest discipline) was missing from the CLAUDE.md role list, README's "13 disciplines" count, and the `AGENTS_REFERENCE.md` role table — added in all three.
- **Counts:** test count `585 → 621`; disciplines `13 → 14`; added `impl_loop.py` and `config/implementation_loop.toml` to the README project structure (were shipped but undocumented).
- **`update --force`** (the clobber-guard bypass from #41) documented in `API.md`, `CLAUDE.md`, and `README.md`.
- **Duplicate** `CLAUDE_CODE_USAGE.md` reference removed in `INTEGRATION_GUIDE.md`.
- **`TICKET-slack-n8n-digest.md`** marked shipped but had all acceptance-criteria boxes unchecked — checked them.

Memory files (MEMORY.md status line, slash-command list, implementation-loop phase status) were also refreshed locally; not in this PR.

## Not changed (verified clean / intentionally preserved)
- All Python docstrings/comments current (agent found zero drift, zero stale TODOs).
- Historical changelog/PR-note counts left as-is (accurate at their version).
- The 4 open stats/n8n-activation tickets are correctly still open.

Docs only — **621 tests still pass**. Stacked on #41 because the `--force` feature it documents lives there; rebase to main once #41 merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #43, opened 2026-06-30
feat(install): extend clobber guard to .claude/ commands & workflows

eladrianovalle:
## Why

PR #41 added a clobber guard: `update` refuses to overwrite installed files that a user has locally edited (drifted from their recorded checksum). But the install **MANIFEST.json only checksummed `.studio/source/` files** — the verbatim `.claude/commands/*` and `.claude/workflows/*` were copied on every install yet never tracked. So a user who locally edited a slash command or the `implementation-loop.js` workflow had that edit **silently overwritten** on the next `update`, with no warning. This closes that altitude gap from #41.

## What

Track the `.claude/` commands & workflows in the same flat manifest:

- **`_claude_manifest_keys()`** — collect keys for the shipped commands + workflows that exist, keyed by their repo-root install path (`.claude/commands/run-phase.md`) to distinguish them from bare source keys (`run_phase.py`, relative to `.studio/source/`).
- **`_manifest_source_path()` / `_manifest_installed_path()`** — location-aware resolution so source files (`.studio/source/…`) and `.claude/` files (`.claude/…`) share one manifest. Keys starting with `.claude/` resolve against repo-root (source) and target-root (install); everything else against `.studio/source/` as before.
- `install_studio` and `check_studio` now feed both key sets into `_build_manifest`; the `locally_modified` scan resolves each key's on-disk path via the helper.

Result: `check-install` reports a locally-edited command/workflow, and `update` blocks on it (unless `--force`) exactly like a source file.

## Tests (+3, 624 total green)

- `test_manifest_covers_claude_commands_and_workflows` — manifest contains `.claude/commands/*` + `.claude/workflows/*` keys with correct checksums
- `test_check_reports_locally_modified_command` — edited command shows up in the clobber set
- `test_update_blocked_on_locally_modified_command` — update blocks on an edited command; `--force` overwrites it

Verified end-to-end against a real temp install: 8 commands + 1 workflow tracked, workflow edit detected, update blocked, edit survived.

## Notes

- **Transition gap (by design):** existing installs have no recorded baseline for `.claude/` files yet, so the *first* `update` after this lands can't protect a pre-existing command edit — there's no checksum to compare against (same situation source files were in before the manifest existed). After that first reinstall writes the baseline, future edits are protected.
- The user-facing `--force` doc currently says "snapshot files" (in #42, unmerged). Once #42 merges I'll rebase and broaden that wording to "installed files (source + commands/workflows)".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ All notable changes to TheGameStudio are documented here.
 - `No Merge Conflicts` CI workflow (`.github/workflows/no-merge-conflicts.yml`) to satisfy branch ruleset
 
 ### Fixed
+- Clobber guard now covers the verbatim `.claude/` commands & workflows, not just `.studio/source/` files — the install manifest checksums `.claude/commands/*` and `.claude/workflows/*` too, so a locally-edited slash command or workflow is reported by `check-install` and blocks `update` (unless `--force`) instead of being silently overwritten
 - `check-install`/`update` stale-snapshot detection (#20) — run through the installed snapshot, both compared it against its own manifest and always reported "up to date", silently blocking updates. Now resolves the live upstream from `VERSION.source_path`, warning loudly when it can't
 - Installer omitted the `integrations` subpackage — `init`/`update` shipped without `integrations/__init__.py` + `slack_digest.py`, breaking every command in installed projects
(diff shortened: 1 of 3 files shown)


eladrianovalle (commented):
APPROVED
The manifest now tracks the installed .claude commands and workflow files, and the clobber guard resolves those paths correctly for check and update.

Must fix
None.

Consider
None.

PR #45, opened 2026-07-01
Human voice: 'Write Code for Humans' principle + doc/test staleness sweep

eladrianovalle:
Extends the human-voice work on this branch from docs into the code itself, plus an `/unstale` + `/detest` pass.

## Draft — not ready for merge yet
Opened as a draft per CODING_PRINCIPLES §5. Flip to ready when you've reviewed the principle wording.

### `docs(principles)` — new §7 "Write Code for Humans"
- Adds a standalone principle governing the readability of the **code itself**. §6 only covered docs/comments; nothing told agents to favor explicit, descriptive, slightly-verbose code over compact-and-clever, so terse output kept slipping through.
- Resolves a latent conflict: §2 ("200 lines → 50") and the Contrarian Mandate ("default to deletion") were being read as *fewer characters is better*. Both now say **cut concepts, not characters** — remove moving parts, never squeeze readable logic into denser code.
- Lands in `CODING_PRINCIPLES.md` (the source injected into every installing repo's `CLAUDE.md`) and root `CLAUDE.md`, plus a new Contrarian Mandate bullet in `scopes.py` so the editor guards readability in every debate.

### `fix(unstale)` — retire a stale doc + a bogus CLI choice
- `record-metrics --agent` listed `"polish"` (a scope name) as a valid agent while omitting `"implementer"` (the real agent in market/design/tech). Corrected; no test pinned the choices.
- `CROSS_REPO_CHECK_MVI.md` described a `studio check` command that shipped as `check-install`, with output strings the real command doesn't print. Retargeted to actual behavior; fixed the `INDEX.md` label too.

### `test(clarity)` — strengthen a weak assertion
- The challenge-penalty test only checked `score < 0.667` (directional), which would survive deleting the penalty (2/3 rounds just under 0.667). Pinned to the exact `pytest.approx(2/3 - 0.1)`. Mutation-verified: dropping the penalty and scaling `0.1→0.2` both now fail.

## Verification
- 624 tests pass.
- Contrarian Mandate change confirmed reaching the scope prompts.
- Clarity fix confirmed by deliberate mutation (reverted).

## Propagation note
§7 reaches installing repos (Pictorly, cemetery-security) on their next `update`, which rewrites their `CLAUDE.md` principles block via the sentinel injection.


eladrianovalle (commented):
REJECTED
The parser change is fine, but one Studio command template still tells users to pass the removed agent value, so that workflow can now fail.

Must fix
- `.claude/commands/run-studio-phase.md:40`: this template still documents `--agent {advocate|contrarian|integrator|polish}`, but `studio/run_phase.py` now accepts `{advocate,contrarian,integrator,implementer}`. The generated instruction tells Studio runs to record the polish/implementer agent with `--agent polish`; that now exits with `invalid choice: 'polish'`, so update this template. `studio/docs/API.md:96` has the same stale list and should be brought in sync too.

Consider
- None.

eladrianovalle (commented):
REJECTED
The API doc was brought in sync, but the generated Studio slash command still tells users to pass the removed `polish` agent value.

Must fix
- `.claude/commands/run-studio-phase.md:40`: this template still documents `--agent {advocate|contrarian|integrator|polish}`, while `studio/run_phase.py:1884` now accepts `{advocate,contrarian,integrator,implementer}`. A user following the installed command will still record the polish/implementer step with `--agent polish`, which now exits with `invalid choice: 'polish'`.

Consider
- None.

eladrianovalle (commented):
APPROVED
The follow-up fixes the stale Studio command template and brings the agent choices back in line with the parser and API docs.

Must fix
- None.

Consider
- None.

PR #54, opened 2026-07-03
fix: /studio-update refreshes the CLAUDE.md principles block

eladrianovalle:
## The bug

Studio's coding principles get dropped into a project's `CLAUDE.md` **once, at install time**, between two marker comments. `check-install` compares *source file* checksums — and `CLAUDE.md` isn't a source file. So when the principles moved ahead upstream (the new **Talk to Humans** and **Write Code for Humans** principles), the update had nothing tracked that changed: it said "up to date" and never refreshed the block. The block silently fell behind.

## The fix

- `check-install` now compares the installed principles block against the current template and flags it (`claude_md_stale`), so it registers as an available update.
- `update` re-injects the block in place and reports it (`claude_md_refreshed`), instead of no-oping.
- **Your own notes are preserved** — the existing injection only touches text between the markers; everything above and below is left exactly as-is. A `CLAUDE.md` you deliberately deleted is not resurrected by a routine check.

## Verification

- 6 new tests (detection, missing block, refresh-preserving-notes, update-runs-when-only-block-stale, deliberately-removed-file).
- Full suite: **688 passed**; ruff clean.
- Drove the real CLI end-to-end: fresh install → up to date; drift the block → check flags `CLAUDE.md: coding-principles block is behind the current template`; update → `refreshed the coding-principles block (your own notes left untouched)`; block restored, drift gone, notes intact; second update → genuinely up to date.

## Note

Built in a dedicated worktree so the source checkout stayed on `main` the whole time — the discipline we just agreed on, so no work-in-progress leaks into other projects' update checks.


eladrianovalle (commented):
APPROVED
Clean CLAUDE.md sync: \`_principles_block_stale\` detects when the injected block has drifted from the template, \`check_studio\` folds it into \`up_to_date\`, and \`update_studio\` re-injects via the existing \`install_studio\` path. Six targeted tests cover the key scenarios.

**Must fix** — none.

**Consider:**
- \`install.py:310\` — if \`_SENTINEL_END\` appears before \`_SENTINEL_BEGIN\` in a hand-edited file, \`content[start:end]\` is empty and the function returns \`True\` (false positive, harmless extra refresh). A \`if start >= end: return True\` guard makes the intent explicit.
- \`tests/test_install.py:506\` — \`_drift_block\` replaces the literal string \`"Talk to Humans"\` to manufacture drift; if that principle is renamed upstream the helper silently no-ops and the drift-detection tests would fail loudly on the next run. Appending a fixed byte to the block content would be template-agnostic.

PR #55, opened 2026-07-04
fix: check/update read Studio source from main, not a parked branch

eladrianovalle:
## The problem

Other projects don't pull Studio from GitHub — they resolve their upstream to a **local working tree** and read its files directly. So when this source repo was left parked on a feature branch, that work-in-progress leaked into other projects' update checks: they compared against the branch, and once they'd copied it, reported a false "up to date."

## The backstop

`check-install`/`update` now read Studio source from the **default branch** (`main`) regardless of what's checked out:

- Materializes `main`'s *committed* tree in a throwaway detached git worktree, reads/copies from there, then cleans it up.
- Prints a note naming the branch it bypassed (e.g. `read Studio source from 'main' (the source checkout is on 'feature-x')`).
- `update` copies from the materialized tree but records the durable upstream path in `VERSION` — never the temp worktree — via a new `install_studio(source_path_override=…)`.
- **Fast path:** when the source is already on `main` and clean, it's used directly — no worktree churn. Falls back to the working tree when the source isn't a git repo or the default branch can't be resolved.

Belt-and-suspenders on top of the "work in a worktree, keep main checked out" habit — even if a checkout slips onto a branch, consumers still see `main`.

## Verification

- 5 new helper tests (disabled / non-git / fast-path / reads-main-from-feature-branch incl. worktree cleanup / reads-committed-main-when-dirty).
- Full suite **693 passed**; ruff clean.
- End-to-end: installed a throwaway project from this checkout while it sat on a feature branch; `check-install` printed `read Studio source from 'main'` and compared against `main` instead of falsely reporting up-to-date.

Built in a dedicated worktree — this source folder stayed on `main` throughout.


eladrianovalle (commented):
APPROVED

Solid fix. The `_source_at_default_branch` context manager correctly materializes the committed `main` tree in a throwaway worktree and cleans up in `finally` — `added` guard ensures the removal only runs when `git worktree add` succeeded. Fast-path (already on main+clean) avoids the worktree churn correctly, and the `not _git_out(repo, "status", "--porcelain")` empty-string truthiness works as intended; the `head is not None` guard upstream also prevents a git-failure edge case from misfiring it.

The `source_path_override` mechanism is the subtle-but-critical piece: `update_studio` copies from the throwaway tree but records the durable upstream in `VERSION` so the next run knows where to look. `override = source_dir if effective_dir != source_dir else None` correctly limits this to the worktree case only.

`auto_resolved` flag preventing double-materialization when `update_studio` calls `check_studio(target, effective_dir)` with an explicit dir is correct — the inner check sees `studio_dir is not None` and skips re-entering the worktree.

Tests cover disabled, non-git, fast-path, parked-on-feature-branch (with uncommitted edits and worktree cleanup verification), and dirty-on-main. That's the meaningful behavioral surface.

**Consider:** `note is None` when dirty-on-main (bypassed to committed main but no branch was skipped) is arguably silent — a user who left uncommitted edits on `main` gets no indication the working tree was bypassed. Low-impact since the committed and dirty states should be close, but worth knowing if it ever causes confusion.

eladrianovalle (commented):
APPROVED

Doc-only follow-up: broken-snapshot recovery path and matching changelog entry.

The `studio-update.md` block is correctly scoped — it triggers only on outright run failure (traceback / `ModuleNotFoundError` / `ImportError`), not on a bad report, and directs the reader to the upstream entrypoint rather than trying to patch the broken snapshot in place. The instruction to read `source_path` from `.studio/VERSION` is the right anchor, and the fallback ("tell the user you need the path") handles the missing-VERSION edge cleanly. No ambiguity about which entrypoint to use for Step 3 during recovery.

Changelog entry matches the new behavior faithfully.

PR #56, opened 2026-07-04
docs: /studio-update self-recovers from a broken installed snapshot

eladrianovalle:
## What

When a project's installed Studio copy is too old or broken to run, its `run_phase.py` fails at **import time** — before any of the command's existing WARNING handling can fire. The user got a raw Python traceback and was stuck. (It only recovered in practice because a personal memory note said "run from upstream" — and memory notes don't travel to other people's installs.)

`/studio-update` now recognizes an outright run failure (`ModuleNotFoundError`/`ImportError`/traceback) and switches to running `check-install`/`update` from the **upstream** Studio repo, pointing at the project. The upstream copy is current, and its `update` re-copies a working snapshot in, repairing it — no hand-debugging.

- Reads the upstream path from `.studio/VERSION` → `source_path`; asks the user for it if that's missing.
- Tells the agent to use the upstream entrypoint for the rest of the procedure until the snapshot is repaired.

## Where it ships

- `.claude/commands/studio-update.md` — the command instructions (already in the cross-repo install set, so every install gets it).
- `CHANGELOG.md` — recorded under `[Unreleased] → Fixed`.

Docs-only; no code. Built in a dedicated worktree with the source folder kept on `main`.


eladrianovalle (commented):
APPROVED

Docs-only. The recovery section in `studio-update.md` is clear and correct — using the upstream entrypoint to repair a broken snapshot is the right call, and surfacing `source_path` from `.studio/VERSION` as the discovery mechanism is exactly what's recorded there. CHANGELOG entry is accurate.

PR #57, opened 2026-07-06
feat: add /spec command + "Spec Before Build" principle

eladrianovalle:
## What & why

Adds **`/spec`**, the front door to a feature: it maps the feature's architecture into an approved, source-of-truth document *before* any code gets written. Once approved, that document is what the feature is built against.

It also enshrines the practice as a durable work-process standard so this isn't a per-task ask.

## How `/spec` works

Reuses the existing debate machinery — **no new Python**, the command markdown drives it:

1. **Resolve identity** — slug, ticket, and the tracked path the spec lives at (`specs/`, or `.studio/specs/` in a consuming repo).
2. **Discovery-forward debate** — a single advocate/contrarian `tech`-phase pass, scope-paced so it front-loads discovery:
   - **Alignment scope** folds in the Open-Questions Pre-Flight to surface and answer the architectural unknowns, pausing on P0 blockers.
   - **Depth scope** pressure-tests the design (edge cases, simpler alternatives, cut scope) and finalizes it.
3. **Synthesize the spec doc** — one document written for two readers at once: a plain-language summary a non-engineer follows, plus build-ready technical detail (components, data flow, interfaces, data model). Carries a **required Mermaid diagram**, the key decisions and cut scope from the debate, open risks, and a build plan of MVI units that bridges to `/studio-implement`.
4. **Approval is the gate** — `status` stays `draft` until the user approves, then flips to `approved`. That's what makes it the source of truth. Linked to its ticket via frontmatter.

Default `--max-iterations 2` = one alignment (discovery) pass + one depth (finalize) pass. Args: feature text, `--ticket`, `--id`, `--max-iterations`, `--plan`.

## Coding Principle 9 — Spec Before Build

A non-trivial feature gets an approved architecture spec before implementation, tracked with its ticket. Explicitly scoped to features and meaningful changes, not one-line fixes. Reaches installing repos through the same `CLAUDE.md` sentinel injection that carries the other principles.

## Changes

- `.claude/commands/spec.md` — the new command (AI-read, like `/smoke` and `/studio-implement`)
- `studio/install.py` — ships `/spec` on cross-repo install
- `studio/docs/CODING_PRINCIPLES.md` + root `CLAUDE.md` — Principle 9
- `README.md`, `CLAUDE_CODE_USAGE.md`, `CHANGELOG.md` — documented (docs contract)

## Verification

- Full suite green: **688 passed**
- `ruff check install.py` clean
- Confirmed `prepare --phase tech … --json` returns the run dir the command relies on

## One open call for the reviewer

In a **consuming** repo, specs land in `.studio/specs/`. If that repo's `.gitignore` blanket-ignores `.studio/`, the spec won't actually be tracked — defeating "source of truth." The command notes specs are meant to be committed, but does not touch any consuming repo's ignore rules. If we'd rather specs always live in a plain top-level `specs/` even in installed repos (guaranteed tracked), that's a one-line change to the command.


eladrianovalle (commented):
APPROVED

Clean addition of the `/spec` command and Principle 9 — well-scoped and consistent with the existing debate machinery.

**Consider**

- `spec.md` embeds a triple-backtick Mermaid example inside a larger triple-backtick block (the spec template section). In most renderers the inner fence terminates the outer one early, breaking the template display. Consider using four backticks for the outer fence or switching the inner Mermaid example to an indented code block to avoid the nesting ambiguity.
- The `--max-iterations` default of 2 is well-calibrated for a focused pass, but the doc only explains this as "one unknowns-surfacing pass + one finalize pass." A line like "raise this only for genuinely large or contested designs" would help users avoid reflex-bumping it.

PR #58, opened 2026-07-06
fix: /spec template's inner mermaid fence breaks the outer code block

eladrianovalle:
## The bug

In `.claude/commands/spec.md`, the spec-template section wraps a sample document in a ` ```markdown ` fence, and that sample contains a ` ```mermaid ` diagram block. Most Markdown renderers close a fenced block at the **first** inner ` ``` ` line — so the outer fence terminates early at the mermaid block, and the second half of the template (everything from the diagram onward) spills out as broken plain text.

## The fix

Bump the outer template fence to **four backticks**. Per CommonMark, a fence is only closed by a line with the same or greater number of backticks, so the inner three-backtick mermaid block no longer ends it. Added a one-line note that the actual spec file `/spec` writes still uses a normal three-backtick fence.

Docs-only change to one command file. No code, no tests affected.


eladrianovalle (commented):
APPROVED
Correct fix for the nested fence collision — upgrading the outer fence to four backticks so the inner ```mermaid block doesn't close it early, with a clear inline note explaining why.

No issues.

PR #59, opened 2026-07-06
spec: source-staleness detection for /studio-update

eladrianovalle:
Approved architecture spec produced via `/spec` (its first real dogfood run). This is the **spec only** — the source of truth to build the fix against, per Coding Principle 9. Implementation follows.

## The problem

`/studio-update` measures "up to date" against your **local** Studio source folder. It reads that folder's committed `main`, prefers local `main` over `origin/main`, never fetches, and content-compares via sha256 — so if your Studio checkout is behind its remote (unpulled), the files look byte-identical to what your project has and it confidently reports **"up to date"** against a stale mirror. This is the recurring pain in consumer repos (a consumer couldn't get `/spec` no matter how many times they updated).

## The design

Before comparing, a new best-effort helper checks whether the resolved source is itself behind its remote: a bounded `git fetch` of the default-branch remote (~5s timeout, `--no-fetch` opt-out, offline-safe fallback to cached refs), then `rev-list --left-right --count`. When local `main` is **behind** `origin/main`:
- suppress the "up to date" verdict and `sys.exit(1)` (no false green),
- materialize and diff against `origin/main` so the user sees the honest update list,
- print the exact `git -C <source> pull` command.

**Never mutates the user's source repo** (no auto-pull; `--pull-source` deferred). Every can't-tell case — offline, no remote, detached HEAD, not-a-repo, or local merely *ahead* (unpushed WIP) — falls through to today's behavior. It only ever adds a block, never a false one.

## What the debate caught (pre-build)

The tech-phase contrarian **rejected** the first design over a verified correctness blocker: it routed staleness through `check_studio`, but `update_studio` calls `check_studio` with an explicit dir → `enabled=False` there → staleness would have **silently never run on `/studio-update`**, the exact bugged command. Confirmed against `install.py`. The corrected design computes staleness in each handler's own scope, with `origin/main` materialization threaded through update's path end to end. That correction is the load-bearing decision in the spec.

## Scope

Source-behind-remote detection only. The second staleness layer (a consumer's vendored snapshot too old to self-heal) is a **noted follow-up** — it already has recovery guidance in the update command.

## Build plan (for `/studio-implement`)

1. Detection helper `_source_staleness` + `SourceStaleness` + `_git_fetch` + `_default_branch_ref_local`, unit-tested against temp git repos.
2. Wire into `check-install` (verdict + exit code + `--no-fetch`).
3. Wire into `update` in its own scope (regression test guards the delegation trap).
4. Docs: `studio-update.md`, arg docs, CHANGELOG.

Debate record: `.studio/output/tech/run_tech_20260706_193131` (design REJECTED at v1 → corrected → spec APPROVED). Full spec with Mermaid diagram: `specs/update-staleness-detection.md`.

### DIFF
--- a/specs/update-staleness-detection.md
+++ b/specs/update-staleness-detection.md
@@ -1,0 +1,235 @@
+---
+feature: Source-Staleness Detection for /studio-update
+slug: update-staleness-detection
+ticket: none
+status: approved
+studio_run: .studio/output/tech/run_tech_20260706_193131
+---
+
+# Source-Staleness Detection for `/studio-update` — Architecture Spec
+
+## In Plain Language
+
+When you run `/studio-update` in a project, it checks your installed Studio copy against
+the "real" Studio source and tells you whether you're up to date. The problem: it measures
+"up to date" against **your local Studio folder**, and it never checks whether that folder is
+itself behind. If your Studio checkout hasn't pulled the latest commits, the files it compares
+look identical to what your project already has — so the tool confidently says "up to date"
+when it isn't. You then can't get new commands like `/spec` no matter how many times you update.
+
+This feature teaches the updater to first ask a second question: **"is the source I'm about to
+compare against itself behind its own remote?"** It does a quick network fetch of the Studio
+source's remote (with a short timeout, and a `--no-fetch` escape hatch for offline use), and if
+your local `main` is behind `origin/main`, it refuses to print "up to date." Instead it compares
+against the fresher `origin/main`, shows you what's actually new, and tells you the one command to
+run to catch your source up. A false "up to date" over a stale source becomes impossible.
+
+It never touches your Studio checkout for you — no silent `git pull`. It just tells you the exact
+command. And it degrades quietly: offline, no remote, or a source that isn't a git repo all fall
+back to today's behavior rather than crashing.
+
+## Architecture at a Glance
+
+```mermaid
+flowchart TD
+    subgraph cmd["check-install / update"]
+      Resolve["resolve source dir<br/>(_resolve_source_dir)"]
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
APPROVED

Spec-only PR adding a well-grounded design for source-staleness detection in `/studio-update`; I verified its load-bearing code claims against the current tree and they all hold.

**Must fix** — none.

Verified claims:
- `update_studio` (install.py:670) does call `check_studio(target, effective_dir)` with an explicit dir, so `check_studio`'s `enabled` (install.py:608) is indeed `False` in that path — Key Decision #2 (compute staleness in update's own scope) is real and necessary, not hypothetical.
- `_source_at_default_branch`'s clean-tree fast path (install.py:271) would materialize a stale local `main` after a fetch; the spec's "pass the target ref, skip the fast path" contract addresses it correctly.
- `except Exception` at run_phase.py:3284 does not catch `SystemExit`, so the handler-level `sys.exit(1)` propagates as designed.
- Referenced anchors (subparsers ~2042/2052, handlers ~3060/3094) match.

**Consider**
- Exit-code asymmetry: today `check-install` exits 0 even when "needs updating" (run_phase.py:3058); after this spec, stale-source exits 1 but ordinary out-of-date still exits 0. Worth deciding deliberately whether plain "needs updating" should also be non-zero, or documenting the split, so CI consumers get a coherent contract.
- `_default_branch_ref_local` keys on `main`/`master` locally; a source repo with local `master` but a fresher `origin/main` would compare `master...origin/master` and miss it. Covered by the acknowledged default-branch assumption, but a one-line note in the spec's risk section would make it explicit.

PR #60, opened 2026-07-07
feat: source-staleness detection for check-install/update

eladrianovalle:
Implements the approved spec `specs/update-staleness-detection.md` (spec PR #59, now merged). Built to the spec's four-unit plan via `/studio-implement` (writer/editor loop), each unit gated on tests-green.

## What this fixes

`/studio-update` measured "up to date" against your **local** Studio source checkout — reading its committed `main`, never fetching, comparing file contents — so if that checkout was behind its remote (unpulled), everything looked byte-identical and it confidently reported **"up to date"** against a stale mirror. The recurring pain in consumer repos (a consumer couldn't get `/spec` no matter how many times they updated).

## What changed

A bounded, best-effort `git fetch` of the source's default-branch remote (short timeout, `--no-fetch` opt-out, offline-safe fallback) now checks whether local `main` is behind `origin/main`:
- **`check-install`** suppresses the "up to date" verdict, exits non-zero, compares against `origin/main`, and prints the exact `git -C <source> pull` command.
- **`update`** refuses to no-op and reinstalls from `origin/main` instead.
- Never mutates your source repo (no auto-pull). Every can't-tell case (offline, no remote, detached HEAD, not-a-repo, or local merely *ahead*) falls through to today's behavior — only ever adds a block, never a false one.

## How it was built (4 units, writer/editor loop)

1. **Detection helper** — `_source_staleness` + `SourceStaleness` + `_git_fetch` + `_default_branch_ref_local` in `install.py`, unit-tested against hermetic temp git repos (`eccb374`).
2. **check-install wiring** — fold `is_stale` into the verdict, `origin/main` materialization, `--no-fetch`, handler `exit(1)` + regression guard (`8010e12`).
3. **update wiring** — staleness computed in `update_studio`'s **own scope**, reinstall from `origin/main` (`24e47d5`).
4. **Docs** — `studio-update.md`, `API.md`, CHANGELOG (`37f2edf`).

## The load-bearing correction

The design debate's contrarian caught that routing staleness through the delegated `check_studio` would silently disable it on `/studio-update` (it's called with an explicit dir → `enabled=False`). The implementation computes staleness in each handler's own scope, and `test_update_stale_source_reinstalls_from_origin` **permanently guards the trap** — it fails if anyone reroutes it through the delegate.

## Tests

Full suite **712 passing** (was 688 on `main`; +24 new). Each unit's tests, the delegation-trap guard, the clean-source regression guards, and AI-TDD mutation checks all green. `main` was never touched during the build (isolated worktree).

## Out of scope (noted follow-up)

The second staleness layer — a consumer's vendored snapshot too old to self-heal — is deliberately not here; it already has recovery guidance in `studio-update.md`.


eladrianovalle (commented):
APPROVED

Solid, well-guarded fix: staleness detection is computed independently in each handler's own scope, strictly best-effort (every can't-tell case falls back to today's behavior), and the ahead-only / offline / no-remote / detached-HEAD edges are all covered by hermetic tests against a local bare remote.

**Must fix**

None.

**Consider**

- `studio/install.py:141` — `_source_staleness` checks for `refs/remotes/origin/<branch>` *before* fetching, so a source with a remote configured but never fetched (remote added by hand, no tracking ref yet) bows out as "origin not configured" even though the fetch would have created the ref. Rare in practice (clones always have it), but reordering fetch-then-check would catch it.
- `studio/install.py:335` — `check_studio` now returns a `SourceStaleness` dataclass inside its dict; any caller that JSON-serializes the status (now or later) will choke on it. A `dataclasses.asdict()` at the boundary would keep the return plain-data.
- `studio/run_phase.py:3092` — the stale-source block in `_do_check_install` exits before the LOCAL EDITS clobber preview prints, so a user with local edits over a stale source first learns about the block only when `update` refuses. Safe (update still blocks), just one extra round-trip.

eladrianovalle (commented):
APPROVED

Follow-up on 37f2edf..1c88319: all three Consider items from the previous review are addressed, each guarded by a new hermetic test, and no new issues in the net-new code.

**Must fix**

None.

**Consider**

- Nothing new. The fetch-before-ref-check reorder correctly handles the never-fetched-remote case (and the offline fallback to cached refs still works via `fetched=False` flowing into the bow-out reason); `asdict()` at the `check_studio`/`update_studio` boundary is applied on all three return paths with handlers consistently switched to dict access; and `_print_local_edits_preview` keeps the clobber warning visible on the stale-source early exit without duplicating the print logic.

PR #61, opened 2026-07-08
Rename /studio-implement slash command to /forge

eladrianovalle:
Renames the writer/editor implementation-loop command from `/studio-implement` to `/forge`.

## What changed
- Command file `.claude/commands/studio-implement.md` → `.claude/commands/forge.md` (title updated to "Forge").
- The shipped-command list in `install.py`, the install test assertion, and every live doc/cross-ref (`README`, `CLAUDE.md`, `CLAUDE_CODE_USAGE.md`, `IMPLEMENTATION_LOOP_SPEC.md`, `INDEX.md`, `SESSION_ANALYTICS_PLAN.md`, `spec.md`, code comments) follow.
- CHANGELOG entry under Changed.

## What did NOT change
The underlying loop machinery keeps its name: `impl_loop.py`, `.claude/workflows/implementation-loop.js`, `IMPLEMENTATION_LOOP_SPEC.md`. The dated `FABLE5_REVIEW_2026-07-01.md` snapshot is left as historical record.

Self-contained and green (install + impl_loop suites pass). Independent of the update-nudge feature PR.


eladrianovalle (commented):
APPROVED

Clean mechanical rename of `/studio-implement` to `/forge` — the command file, install list, workflow comment, tests, and live docs all move together, and the underlying loop machinery deliberately keeps its name.

**Must fix**
None.

**Consider**
- `update_studio` never removes files (`install.py` — `"removed": 0  # We don't remove extra files`), so a consuming repo that runs `/studio-update` after this lands gets `forge.md` installed but keeps a fully working `studio-implement.md`: two live copies of the same command, with the old one dropped from the new manifest so it lingers as an unmanaged orphan forever. A one-off retired-files cleanup list in install (or at least a CHANGELOG/usage note telling consumers to delete the old command file) would close that.

PR #62, opened 2026-07-08
Proactive update-availability nudge for consuming repos

eladrianovalle:
Consuming repos learn — once, at session start — that their installed Studio is behind upstream, and get nudged to run `/studio-update`, without having to remember to check. Built to the approved spec (`specs/update-availability-nudge.md`) via `/spec` → `/forge`.

## How it works
`init`/`update` install a per-user Claude Code **SessionStart hook** (in `.claude/settings.local.json`) that runs a new lightweight `check-updates` command. It compares the commit you installed from against the source's current `origin/main` HEAD; if they differ, it surfaces a one-line nudge, silent otherwise.

- **Cheap & quiet:** bounded best-effort `git fetch` at most once/24h (cached in `.studio/update-check.json`), near-instant on repeat sessions, offline-safe, never mutates files, **always exits 0** so it can't break a session.
- **Notify once per new upstream commit.**
- **Opt out:** `--no-hook` on init/update, or a durable empty `.studio/update-check.off`.
- **Machine-local by design:** hook uses the absolute `sys.executable` and lives in the gitignored `settings.local.json` — it only fires on the box that ran `studio init`; a teammate who just cloned the repo gets a safe no-op.

## Built in 3 MVI units (via /forge)
1. `compute_update_check` + `UpdateCheck` + 3-field cache + `check-updates` command.
2. The SessionStart hook install (merge-not-clobber, idempotent, `--no-hook`, opt-out).
3. Docs (API, usage guide, bridge template, README, CHANGELOG).

## Verification
Full suite 730 green; new `tests/test_update_check.py` (8) + `tests/test_hook_install.py`. Live demo confirmed: silent when current, emits the exact SessionStart nudge when the source is ahead.

## Open gate
The spec's top risk — whether the nudge line actually reaches the user on screen vs. only the model's context — needs a real Claude Code session in a consuming repo to confirm before merge. The emitted format matches Claude Code's SessionStart `additionalContext` contract.

## Note
References `/forge` in docs; best merged after the rename PR (`chore/rename-forge`). Independently mergeable — no code dependency on the rename.


eladrianovalle (commented):
REJECTED

Well-built feature — hermetic tests, sensible spec, and the never-break-a-session posture is right — but there are two contract-level gaps in exactly the surfaces the feature exists for.

**Must fix**
- `studio/install.py:512` (`_hook_command`) — the hook command references the script by a cwd-relative path (`".studio/source/run_phase.py"`) while passing `$CLAUDE_PROJECT_DIR` only as `--target`. Claude Code runs hook commands from the session's working directory, not necessarily the project root: launch `claude` from a subdirectory of the consuming repo and Python exits 2 (`can't open file`), surfacing a hook error at session start — breaking the always-exit-0 contract the whole design leans on. Anchor the script path too: `"$CLAUDE_PROJECT_DIR/.studio/source/run_phase.py"`. While there, quote `{sys.executable}` — an interpreter path containing a space breaks the command as written.
- `studio/install.py:1139` (`update_studio`) — `update --no-hook` is documented (API.md, CLAUDE_CODE_USAGE.md) to *remove* the hook, but both early returns in `update_studio` (already-up-to-date, and blocked-on-local-edits) exit before `install_studio` runs, so `install_hook` is never consulted. The most common opt-out scenario — user is current and runs `studio update --no-hook` to get rid of the nudge — silently no-ops and the hook stays. Handle the hook (install or remove) before the up-to-date short-circuit, or call `_install_sessionstart_hook` on the early-return paths.

**Consider**
- `compute_update_check`'s "NEVER raises" docstring isn't quite true: a cache whose `last_check` is a non-number (hand-edited or corrupted) makes `now - cache.get("last_check", 0)` raise `TypeError` on every run. The handler swallows it, but the feature is then permanently dead until the file is deleted, with no path back. Cheap to type-check the three fields in `_load_update_cache`.
- `_install_sessionstart_hook` guards a non-object settings file but not non-object *fields*: an existing `"hooks": []` or `"SessionStart": {}` raises `AttributeError` out of `install_studio`, despite the never-raises docstring.
- The merge gate you set yourself in the spec — `/smoke` proof that the nudge actually reaches a live session — still stands; nothing in this PR discharges it.


---
🤖 reviewed by `claude-fable-5` — stamped by pr-gate from the runtime's agent config (not self-reported).

eladrianovalle:
Thanks — both must-fixes and both robustness gaps were real and are now addressed (rebased onto current `main` post-rename; conflict resolved). Full suite **737 green**.

**Must fix**
- **`_hook_command` cwd-relative path** — now anchors the script to `"$CLAUDE_PROJECT_DIR/.studio/source/run_phase.py"` and quotes the interpreter. Verified by running the exact emitted command from a subdirectory of a fresh install: it resolves and exits 0 (previously exit 2). Test: `TestHookCommand.test_command_paths_are_absolute_and_quoted`.
- **`update --no-hook` skipped on early returns** — `update_studio` now (un)installs the hook **up front**, before both short-circuits (up-to-date and blocked-on-local-edits); the re-install passes `install_hook=False` so it isn't done twice. Regression test: `TestUpdateHookOnEarlyReturn.test_update_no_hook_removes_hook_when_up_to_date` (+ a refresh-on-up-to-date test).

**Consider (both fixed)**
- **Corrupt cache raising** — `_load_update_cache` discards a cache whose `last_check` isn't numeric (or whose commit fields are an outright wrong type), so the TTL math can't raise and permanently wedge the check. Tests in `test_update_check.py`. (`notified_commit: None` stays valid.)
- **Non-object hook fields** — `_install_sessionstart_hook` now guards a non-object `hooks` / non-list `SessionStart` the same way it already guarded a non-object settings file. Tests: `TestHookMalformedFields`.

**Still open, as you noted:** the `/smoke` proof that the nudge actually reaches a live session is not discharged by this PR. Leaving it as the pre-merge gate — it needs a real Claude Code session in a consuming repo on this machine.

eladrianovalle (commented):
APPROVED

Both must-fix items and both consider items from the prior review are fixed, each backed by a new regression test; full suite is green (737 passed).

**Verified**
- `studio/install.py` `_hook_command` — script path is now anchored to `"$CLAUDE_PROJECT_DIR/.studio/source/run_phase.py"` and the interpreter is quoted. Covered by `TestHookCommand.test_command_paths_are_absolute_and_quoted`.
- `update_studio` — `_install_sessionstart_hook` now runs up front, before both early-return short-circuits (up-to-date, blocked-on-local-edits), with the re-install passing `install_hook=False` so it isn't done twice. Covered by `TestUpdateHookOnEarlyReturn` (removes on `--no-hook`, refreshes on a normal update).
- `_load_update_cache` now discards a cache with a non-numeric `last_check` or non-str/None commit fields instead of letting the TTL math raise later. Covered by `test_load_update_cache_discards_bad_last_check` / `_discards_non_string_commit`.
- `_install_sessionstart_hook` now guards non-object `hooks` and non-list `SessionStart` the same way it already guarded a non-object settings file, warning and leaving the file untouched instead of raising. Covered by `TestHookMalformedFields`.

Rest of the diff since my last review is just the `/studio-implement` → `/forge` rename pulled in by the rebase onto main — no functional content there.

**Consider**
- The `/smoke` live-session proof noted last round is still open per your own comment; fine to treat as the pre-merge gate rather than blocking this review.


---
🤖 reviewed by `claude-sonnet-5` — stamped by pr-gate from the runtime's agent config (not self-reported).

PR #63, opened 2026-07-09
Opt-in source fast-forward for /studio-update (stop the recurring 'go pull your source' nag)

eladrianovalle:
Kills the recurring consumer-repo friction: if you develop Studio in one repo and use it from others, you merge PRs (the remote moves ahead) but never `git pull` the local source checkout every consumer reads from — so every `/studio-update` reinstalls from `origin/main` and nags you to pull it by hand, forever. Built to the approved spec (`specs/pull-source-opt-in.md`) via `/spec` → `/forge`.

## What it does
Opt in and `update` fast-forwards your source checkout *for* you when it's cleanly behind:
- **Set once** (covers every consumer on the machine): `[update] auto_pull_source = true` in the **source** repo's `.studio/update.toml`.
- **Or one-off:** `studio update --pull-source`.

After a successful ff the source is genuinely current, so the update is a clean no-op with no nag.

## Safety (it mutates a git repo, so this is the crux)
**Fast-forward only**, and only when the source is: clean working tree, on its default branch, and simply behind origin. Anything off — dirty, diverged, detached, or on a feature branch — changes nothing and falls back to today's reinstall-from-origin + manual-pull hint. Never a merge, force, or rebase. The `HEAD == default branch` precondition is the load-bearing guard (staleness is measured on the local branch regardless of where HEAD sits, so a blind ff could otherwise move the wrong ref). `check-install` stays read-only. Reverses the earlier "never mutate your source" default (source-staleness Key Decision #5), now behind the opt-in + ff-only envelope.

## Built in units (via /forge)
1+2. `_source_auto_pull_enabled` + `SourcePull` + `_fast_forward_source` (the safe primitives).
3+4. `pull_source` wiring in `update_studio` + `--pull-source` flag/output in `_do_update`.
5. Docs (API, usage guide, README, CHANGELOG).

## Verification
Full suite **755 green**; new `tests/test_pull_source.py`. Live end-to-end confirmed against a genuinely-behind source: `studio update --pull-source` fast-forwarded the source HEAD (`160938e → 0aefa45`, a pure ff — 0 merge commits), left it clean, and suppressed the nag. Negative cases (dirty / feature-branch / diverged) leave HEAD untouched.


eladrianovalle (commented):
APPROVED

The opt-in source fast-forward path is conservative, covered by targeted git-backed tests, and preserves the existing stale-source fallback when the pull is unavailable.

Must fix
None.

Consider
The update_studio docstring still says the source repo itself is never mutated; refresh that wording so it reflects the new opt-in behavior.

---

PR #64, opened 2026-07-11
feat(impl-loop): optional model override for writer + editor agents

eladrianovalle:
## What
Adds an optional `model` unit arg to the implementation-loop workflow, passed to `agent()`'s `model` opt on **both** the writer and editor calls. `undefined` inherits the session model (prior behavior), so it's a **no-op unless a caller sets it**.

## Why
The writer/editor agents previously always inherited the session model, with no way to pin a forge run to a specific model. This enables running the loop on a chosen model (e.g. `model: 'fable'`) — surfaced while forging a large port and wanting the whole loop on Fable 5.

## Change
- `implementation-loop.js`: `model: unit.model` on the writer (`schema: WRITER_HANDOFF`) and editor (`schema: EDITOR_HANDOFF`) `agent()` calls.
- Documented the knob on `DEFAULT_UNIT` (`model: undefined`).
- `node --check` passes.

## Follow-up (not in this PR)
The `/forge` skill constructs the Workflow args and doesn't yet surface a `--model` flag, so today `model` is set by invoking the Workflow tool directly. A small `/forge --model <name>` passthrough would make it first-class.

### DIFF
--- a/.claude/workflows/implementation-loop.js
+++ b/.claude/workflows/implementation-loop.js
@@ -14,6 +14,10 @@ export const meta = {
 // ---------------------------------------------------------------------------
 const DEFAULT_UNIT = {
   unit_id: 'unit_impl_loop_config',
+  // Optional model override for the writer + editor agents (e.g. 'fable', 'sonnet',
+  // 'opus'). Passed straight to agent()'s `model` opt; undefined inherits the
+  // session model (the prior behavior), so this is a no-op unless callers set it.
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
APPROVED
The workflow now cleanly supports an optional model override for the writer and editor agents without changing the default inherited-model behavior.

Must fix
None.

Consider
None.


---

PR #65, opened 2026-07-16
feat: borrow two critique mechanisms from gstack

eladrianovalle:
Two critique mechanisms borrowed from a study of [garrytan/gstack](https://github.com/garrytan/gstack), a ~60-skill Claude Code "virtual engineering team." The full comparison and the roadmap for the rest live in the new `studio/docs/GSTACK_COMPARISON.md`.

## What's here

**1. Quote-to-promote + confidence bands** (`studio/scopes.py`, `CONTRARIAN_MANDATE`)
The contrarian must now quote the exact thing it critiques or mark the finding `(unverified)`, and every finding carries a confidence that sets how loudly it's raised (high → stated plainly; medium → "verify this"; low/unverified → a separate notes list, or dropped unless it would be fatal). This is gstack's cheapest anti-hallucination gate, adapted to our critic role. Flows to every deliverable contrarian prompt; suppressed in `--mode questions`.

**2. Reviewer Concerns in `/forge`** (`.claude/workflows/implementation-loop.js`, spec, `forge.md`)
The writer/editor loop is one-way by design (no re-run debate — that's the convergence guard, already structural). The missing half: a real critique the editor can't act on this pass (breaks green, load-bearing, out of unit scope) used to vanish on revert. The editor now records these in `unresolved_concerns` and writes a `reviewer-concerns.md` artifact, so the second agent's most valuable output survives.

## Verification
- Full suite **755 passed**; `ruff` clean; workflow JS parses.
- Existing `CONTRARIAN_MANDATE` tests still green (they assert membership, so the added bullets are covered).

## Notes
- Docs updated per the documentation contract (`IMPLEMENTATION_LOOP_SPEC.md` + `forge.md`).
- Found (not touched): `test_decision_points.py::TestInstructionDocDecisionSection` fails only in a *partial* test selection due to cwd pollution between test files — pre-existing, invisible in the full-suite run.
- Roadmap R1–R8 for the remaining gstack borrowings is in `GSTACK_COMPARISON.md`; R3 + R5 (quick debate-sharpening wins) are next, on this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #66, opened 2026-07-16
feat(R1): independent verifier for contrarian findings

eladrianovalle:
Implements roadmap **R1** — an independent verifier for contrarian findings — from an approved architecture spec. Builds on the evidence gate shipped in #65 (now in `main`).

## The story

The gate in #65 makes the contrarian quote its evidence and self-rate confidence. But the contrarian still grades its own homework. R1 adds a **genuinely independent second opinion**: a fresh agent re-checks each Medium-confidence finding seeing **only the quoted claim + the judging rules — never the contrarian's reasoning** (the anti-anchoring firewall). Agreement promotes the finding; non-confirmation demotes it.

The `/spec` debate **rejected the naive scope** and reshaped it, which is the point of speccing first:
- An "independent" verifier that's one assistant role-playing both roles is theater → real independence needs a Workflow with separate `agent()` calls.
- Findings are prose today → they need a machine-readable form first.

So it's built as **two sequenced MVI units**.

## What's here

**Unit 1 — structured findings** (`studio/findings.py`): a `FINDING` block format + parser mirroring `decision_points.py`, `CONTRARIAN_MANDATE` now emits findings, `findings.json` load/save. Usable alone — makes the confidence gate machine-readable. (writer + editor pass; editor collapsed the dataclass and dropped dead imports.)

**Unit 2 — the verifier** (`studio/verifier.py` + `.claude/workflows/finding-verifier.js`): a tested Python core (select Medium findings → apply verdict → write adjusted confidence back to `findings.json`) plus a thin JS Workflow shell that spawns a fresh `agent()` per Medium finding with the quote-only firewall. The split keeps the logic testable without a JS harness.

**Fix** (`f32182a`): ships `finding-verifier.js` to installed repos — a gap the `/forge` editor's Reviewer Concerns surfaced on the Unit 2 run (the mechanism from #65 working on its first outing).

## Verification

- Full suite **793 passed**; ruff clean; both workflow JS files parse.
- Unit 2 mutation check: 20 assertions broken, all caught.
- **Known gap:** the Python core is tested; the JS Workflow shell's live end-to-end run is being smoke-tested separately (results to follow in a comment). Not merge-ready until that's confirmed.

## Not in scope (deliberate, per spec)

Verifying High findings (already self-gated); auto-wiring the verifier into the run loop (standalone `finding-verifier.js` invocation is the MVI); generalizing beyond the one path.

Spec: `specs/contrarian-finding-verifier.md`. Roadmap + comparison: `studio/docs/GSTACK_COMPARISON.md`.

### DIFF
--- a/.claude/workflows/finding-verifier.js
+++ b/.claude/workflows/finding-verifier.js
@@ -52,19 +52,11 @@ const VERIFIER_HANDOFF = {
 // the ONLY thing the per-finding verifier is allowed to see.
 // ---------------------------------------------------------------------------
 phase('Select')
-const SELECT_SNIPPET = [
-  'import json, sys',
-  'from findings import load_findings_json',
-  'from verifier import select_findings_to_verify',
-  'run_dir = sys.argv[1]',
-  'findings = load_findings_json(run_dir)',
-  'rows = [',
-  '    {"index": i, "quote": f.quote}',
-  '    for i, f in enumerate(findings)',
-  '    if f in select_findings_to_verify(findings)',
-  ']',
-  'print(json.dumps(rows))',
-].join('; ')
+// The whole selector is one real Python function (verifier.select_rows_for_run), so this stays a
+// single valid statement-per-`;` line. The earlier form built a multi-line list comprehension and
+// joined it with '; ', which is a SyntaxError — it only ran because the agent hand-fixed it.
+const SELECT_SNIPPET =
+  'import json, sys; from verifier import select_rows_for_run; print(json.dumps(select_rows_for_run(sys.argv[1])))'
 const selected = await agent(
   [
     `List the Medium-confidence findings eligible for verification in the run directory \`${runDir}\`.`,
--- a/.claude/workflows/finding-verifier.js
+++ b/.claude/workflows/finding-verifier.js
@@ -39,7 +39,9 @@ const VERIFIER_HANDOFF = {
   properties: {
     finding_id: { type: 'integer', description: 'index of the finding in the ordered findings.json list' },
     verdict: { type: 'string', enum: ['confirmed', 'unconfirmed', 'uncertain'] },
-    resulting_confidence: { type: 'string', enum: ['high', 'medium', 'low'], description: 'optional; derived from the verdict when omitted' },
+    // No resulting_confidence override: confidence tracks veracity (two voices agree -> promote),
+    // never severity. The write-back always derives it from the verdict — confirmed promotes,
+    // unconfirmed demotes, uncertain leaves it. Whether a real flaw is *minor* is a separate axis.
     one_line_reason: { type: 'string', description: 'one line, from the verifier\'s own read of the code — not the contrarian\'s argument (it never saw it)' },
   },
 }
@@ -111,11 +113,11 @@ for (const item of eligible) {
       `The quote is \`file:line\` — "exact code/text". You MAY open that file at that line and read the`,
       `surrounding code to judge it. Do NOT go looking for any review notes, comments, or argument about it.`,
       ``,
-      `Demotion rules — return a verdict:`,
-      `- "confirmed": you independently see a real problem in this code -> resulting_confidence "high".`,
-      `- "unconfirmed": you read the code and cannot find the alleged problem -> resulting_confidence "low".`,
-      `- "uncertain": the quote (even with the code) doesn't give you enough to judge -> confidence unchanged.`,
-      `You may omit resulting_confidence to let the write-back derive it from the verdict.`,
+      `Return a verdict on whether the flaw is real (this is a veracity judgment, not a severity one —`,
+      `do not weigh how minor or major it is, only whether it is real):`,
+      `- "confirmed": you independently see a real problem in this code.`,
+      `- "unconfirmed": you read the code and cannot find the alleged problem.`,
+      `- "uncertain": the quote (even with the code) doesn't give you enough to judge.`,
       ``,
       `Set finding_id to ${item.index}. Give a one-line reason from YOUR OWN read of the code.`,
     ].join('\n'),
@@ -125,7 +127,6 @@ for (const item of eligible) {
     verdicts.push({
       index: item.index,
       verdict: verdict.verdict,
-      resulting_confidence: verdict.resulting_confidence,
     })
     log(`Finding #${item.index}: ${verdict.verdict}`)
   }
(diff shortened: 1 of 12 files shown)


eladrianovalle (commented):
REJECTED
The pure parser/verifier tests pass, but the verifier workflow cannot select findings because the generated Python command is syntactically invalid.

Must fix
- `.claude/workflows/finding-verifier.js:65` joins a multi-line list comprehension with `; `, producing `rows = [; ...]`; running the shown command fails with `SyntaxError` before any Medium findings can be verified.

Consider
- Add a workflow-level smoke test or move the selector into a real Python helper so this path is covered outside prompts.

Tests: `cd studio && python3 -m pytest tests/test_findings.py tests/test_verifier.py -q` (38 passed).


---

eladrianovalle:
## Live smoke — the verifier Workflow, end to end

Ran `finding-verifier.js` against a crafted `findings.json` (2 Medium findings quoting real repo lines, 1 High that must be skipped).

**Round 1 caught a real bug:** the Select agent used a top-level *array* schema, which the Anthropic API rejects (`custom.input_schema.type` must be `object`). The tested Python core was clean — only the untested JS shell had it, and only against the live API. Fixed in `9802748` (wrap as `{ findings: [...] }`).

**Round 2 passed** — `{ verified: 2, confirmed: 1, demoted: 1, uncertain: 0 }`:

| Finding | Confidence | Verdict | Result | Verifier's independent reason |
|---|---|---|---|---|
| O(n²) membership in the select snippet | medium | confirmed | medium* | "recomputed for every finding … just to do a membership test" |
| "priority map silently drops unknown levels" | medium | **unconfirmed → low** | demoted | "used only via `.get(priority, 99)` which safely handles any unknown key; no defect" |
| (High finding) | high | — | untouched | correctly skipped by select |

The headline result: **a false-positive finding was demoted by an independent read of the code** — the verifier opened `decision_points.py`, saw the `.get(..., 99)` default, and rejected the claim. And the firewall held: both verifiers reasoned in their own words from the code, never echoing the contrarian's argument (which they never saw).

\* The confirmed finding stayed `medium` (not promoted to `high`) because the verify agent used the optional `resulting_confidence` override — it confirmed the flaw is real but judged it minor. `apply_verdict` honored the override correctly. Whether the verifier *should* be able to confirm-but-not-promote is a design nuance under discussion.

eladrianovalle (commented):
REJECTED
The schema wrapper fixes the Anthropic tool-shape error, but the selector command still cannot run because the generated Python is invalid.

Must fix
- `.claude/workflows/finding-verifier.js:65` still joins the multi-line list comprehension with `; `, so the command shown at line 71 expands to `rows = [; ...]` and raises `SyntaxError` before the selector can return any findings.

Consider
- Add a workflow-level smoke test for the generated selector command so the JS shell is covered alongside the pure Python tests.

Tests: `cd studio && python3 -m pytest tests/test_findings.py tests/test_verifier.py -q` (38 passed); direct selector command still fails with `SyntaxError`.


---

eladrianovalle:
Follow-up on the smoke: dropped the verify agent's `resulting_confidence` override so **confirmed always promotes** (Medium→High). Confidence = veracity (two voices agree), not severity — the grain both gstack and Studio use everywhere else. Re-smoke confirms it live: `[medium] confirmed → high`, `[medium] unconfirmed → low`, High skipped. Commit `5ee6d90`.

eladrianovalle (commented):
REJECTED
The confidence-override cleanup is directionally right, but the verifier workflow still cannot reach verification because the selector command remains syntactically invalid.

Must fix
- `.claude/workflows/finding-verifier.js:67` still joins the multi-line Python list comprehension with `; `, so the command shown at line 73 expands to `rows = [; ...]` and raises `SyntaxError` before any Medium findings can be selected.

Consider
- Add a workflow-level smoke test for the generated selector command so the JS orchestration path is covered in addition to the pure Python parser/verifier tests.

Tests: `cd studio && python3 -m pytest tests/test_findings.py tests/test_verifier.py -q` (38 passed); direct selector command still fails with `SyntaxError`.


---

eladrianovalle:
## Fixed — selector rejection + a second latent bug

Good catch, and correct: the Select command was invalid Python. Fixed in `f8e703c` / `5adf29b`, and verified the way the review did — by running the generated command **directly**, not just through the agent (which had been silently hand-repairing it, which is why the smoke passed and masked it).

**Root fix (the reviewer's "Consider"):** moved the selector into a real, gate-tested Python function, `verifier.select_rows_for_run(run_dir)`. The shell now shells out to a single valid one-liner (`... print(json.dumps(select_rows_for_run(sys.argv[1])))`), so there's no multi-line comprehension to mangle.

**Second bug the direct run surfaced:** `load/save_findings_json` did `run_dir / "findings.json"` and raised `TypeError: 'str' / 'str'` — the shell passes `sys.argv[1]`, a `str`, but the pure tests only ever passed `Path`. They now coerce to `Path` at the boundary.

Both generated commands (Select **and** Write-back) now run clean directly:
```
$ python -c '... select_rows_for_run(sys.argv[1]) ...' <run_dir>
[{"index": 0, "quote": "..."}, {"index": 1, "quote": "..."}]   # the 2 Medium; High excluded
```

**Coverage added** (also addressing #65's `reviewerConcerns` follow-up — the JS shells were untested):
- `test_verifier.py`: `select_rows_for_run` incl. the str-path regression.
- `test_workflow_shells.py`: statically scans every `.claude/workflows/*.js` and asserts each `agent()` schema's top-level `type` is `object` — **this would have caught the earlier top-level-array-schema 400 before the live API did.** Plus a `reviewerConcerns` wiring check and a runner for the node:tests.
- `workflow-shells.test.mjs`: extracts `collectReviewerConcerns` from the real shell source and exercises it against real `unresolved_concerns` payloads.

Full suite: **802 passed**.

eladrianovalle (commented):
APPROVED
The follow-up fixes the selector failure and adds useful coverage for the workflow shell paths that were previously only exercised live.

Must fix
- None.

Consider
- `.claude/workflows/implementation-loop.js:263` still says the Node test loads from `.claude/workflows/lib/workflow-shells.test.mjs`, but the file now lives under `.claude/workflows/tests/workflow-shells.test.mjs`; harmless, just stale comment text.

Tests: `cd studio && python3 -m pytest tests/test_findings.py tests/test_verifier.py tests/test_workflow_shells.py -q` (47 passed); `node --test .claude/workflows/tests/workflow-shells.test.mjs` (4 passed); direct selector smoke returned the expected Medium row.


---

eladrianovalle (commented):
APPROVED
The follow-up resolves the stale workflow-shell test path comment without changing runtime behavior.

Must fix
- None.

Consider
- None.

Tests: `cd studio && python3 -m pytest tests/test_workflow_shells.py -q` (5 passed); `node --test .claude/workflows/tests/workflow-shells.test.mjs` (4 passed).

PR #67, opened 2026-07-16
feat(R2): doc-parity tests (reframed from rejected codegen)

eladrianovalle:
Roadmap **R2**, reframed by the `/spec` debate. Started as gstack's doc-codegen; the debate **rejected the codegen premise for Studio** and produced a smaller, non-destructive design instead.

## Why codegen was rejected (the valuable part)

gstack generates its docs from source because *there* the source is the richer thing. In Studio it's **reversed** — the reference docs are *intentionally richer than the code*:
- `check-updates` / `update` in API.md are **hundred-word hand-authored explainers** (TTL caching, opt-out sentinels, exit-0 safety) that exist nowhere in argparse `help=`.

Generating that table from argparse would **delete** the authored prose — drift-*infliction*, not prevention. And the CLI surface is already audited by `/unstale` (`run_phase.py --help`) for free, with no evidence the table ever drifted. Full write-up + why gstack's mechanism doesn't transplant: `specs/doc-parity-tests.md`.

## What ships instead: assert, don't generate

`studio/tests/test_doc_parity.py` — three checks that guard the source-of-truth **names** while leaving every hand-authored sentence untouched:

1. **CLI commands, both directions** — `build_parser()` subcommands ↔ the `Command | Description` table in API.md. A new command with no row *and* a stale row with no command both fail.
2. **ScopeConfig fields** documented in `SCOPES_GUIDE.md`.
3. **LoopConfig fields** documented in `IMPLEMENTATION_LOOP_SPEC.md`.

No `docgen` module, no new CLI, no CI wiring — the checks ride the existing `pytest` run.

## Verification

- All 3 parity checks **green today** (no current drift); full suite **805 passed**.
- Confirmed they bite: an earlier iteration of the config check flagged all 8 LoopConfig fields as "missing" (my scan was backtick-only; the keys live in a TOML block) — fixed to a whole-word presence check.

## Not in scope (deliberate)

No codegen / description generation / cross-repo generation (docs ship as static artifacts, so the install-path trap never applies); no reverse check on config keys (fragile against prose — the forward "is it documented?" catches the common failure).

### DIFF
--- a/specs/doc-parity-tests.md
+++ b/specs/doc-parity-tests.md
@@ -1,0 +1,119 @@
+---
+feature: Doc-parity tests (Studio roadmap R2, reframed)
+slug: doc-parity-tests
+ticket: none (roadmap item R2 — see studio/docs/GSTACK_COMPARISON.md)
+status: approved
+studio_run: .studio/output/tech/run_tech_20260716_181851
+---
+
+# Doc-Parity Tests — Architecture Spec
+
+## In Plain Language
+
+Studio's reference docs list things the code defines: the CLI subcommands, the config knobs. When
+someone adds a command or a config field and forgets to document it (or renames one and leaves a
+stale doc entry), the docs quietly go wrong. That's the doc-rot `/unstale` chases by hand.
+
+R2 started as "generate those docs from the code" — copied from gstack. The `/spec` debate
+**rejected that** for Studio, and the reason is worth recording: gstack generates its docs from
+source because *there* the source is the richer thing. In Studio it's the opposite — the reference
+docs are **intentionally richer than the code**. `check-updates`'s argparse `help=` is one line;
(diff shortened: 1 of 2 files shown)


eladrianovalle (commented):
APPROVED
The new doc-parity tests are scoped to name drift, pass against the current docs, and avoid overwriting the richer hand-authored reference prose.

Must fix
- None.

Consider
- Ruff is not installed in this local environment, so I could not run the static gate here.


---

eladrianovalle:
Note on the static gate: `ruff` passed locally in the authoring environment (`ruff check tests/test_doc_parity.py` → "All checks passed!"), and CI runs `ruff` as the enforced gate (the `lint-and-test` job, before pytest). So the static gate is covered even though it could not be reproduced in a ruff-less local environment. `test_doc_parity.py` is plain stdlib (argparse, dataclasses, re, pathlib) with no lint-risky constructs.

PR #68, opened 2026-07-16
feat(R4): delta-based trend alerts in stats

eladrianovalle:
Roadmap **R4** — brings gstack's best anti-noise monitoring rule to Studio's cross-run `stats` dashboard: **alert on the direction a metric is moving, not an absolute threshold, and only once the regression has persisted across 2+ consecutive runs.** A single-run dip is a blip and never alerts.

## What's here

`detect_trend_alerts(runs)` in `studio/stats.py` — pure, stdlib, takes the same enriched run dicts `aggregate_stats` consumes:
- Orders runs oldest-first by `created_iso`, builds a per-run series for each tracked metric (runs missing a metric drop out).
- Flags a metric only when its newest runs form an unbroken worsening streak of **≥2 consecutive run-over-run steps**, each moving in the bad direction by **≥5% relative** (`_trailing_regression_streak`).
- Metrics tracked (all have a genuine per-run value — no invented data): **human rating** (falling), **tokens/run** (rising), **cost/run** (rising).

Surfaces as a **"Trend Alerts"** block near the top of the dashboard (omitted entirely when nothing is regressing) and as a top-level `trend_alerts` key in `--json`. `show_stats` computes it once and hands it to both paths.

**Deliberately not included:** ship/approval *rates* — those are aggregates, not a per-run series; a windowed rate-trend would be extra machinery inventing structure the data doesn't have. Easy to add later.

## Verification
Full suite **813 passed** (11 new tests: single-dip-no-alert, 2-consecutive-fires, improving-no-alert, recovery-breaks-streak, insufficient-history, tokens-rising, sub-threshold-noise, out-of-order input, render + omit). `ruff` clean.

Built in an isolated worktree; reviewed before push.

### DIFF
--- a/studio/run_phase.py
+++ b/studio/run_phase.py
@@ -89,6 +89,7 @@ from stats import (
     _parse_usage_log,
     _summarize_metrics,
     aggregate_stats,
+    detect_trend_alerts,
     format_stats,
     summarize_outcomes,
     summarize_session_health,
@@ -2859,6 +2860,11 @@ def show_stats(args: argparse.Namespace) -> None:
 
     agg = aggregate_stats(runs)
 
+    # Trend alerts: run-over-run regressions that persist across 2+ consecutive
+    # runs (rating falling, tokens/cost climbing). detect_trend_alerts orders by
+    # created_iso itself, so the raw run list is fine to hand over.
+    trend_alerts = detect_trend_alerts(runs)
+
     # Session health: the automatic finalize records, oldest first so the
     # recent-vs-earlier trend split is chronological. The --phase filter above
     # already narrowed `runs`, so this respects it.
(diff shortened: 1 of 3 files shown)


eladrianovalle (commented):
APPROVED
Delta-based trend alerts are wired into the stats dashboard cleanly, with focused coverage for persistence, noise filtering, ordering, and rendering.

**Must fix**
None.

**Consider**
None.

---

eladrianovalle (commented):
APPROVED
Follow-up review: the new push only brings in doc-parity coverage on top of the already-reviewed trend-alert change, and the added tests plus the trend-alert suite pass locally.

**Must fix**
None.

**Consider**
None.

PR #73, opened 2026-07-16
fix: clear error on out-of-range verdict index in apply_verdicts_to_run

eladrianovalle:
Reported while reviewing a downstream Studio-update PR.

`apply_verdicts_to_run` indexed `findings[index]` straight from each verdict record. A stale or out-of-range index — e.g. if `findings.json` was edited or shrank between selecting verdicts and writing them back — raised a bare `IndexError`, a rougher failure than the rest of the module's `ValueError` style.

**Fix:** bounds-check the index and raise a `ValueError` that names the offending index and points at the likely cause (findings.json changed since selection). Adds a test for the out-of-range case.

All 15 `test_verifier.py` tests pass.

### DIFF
--- a/studio/tests/test_verifier.py
+++ b/studio/tests/test_verifier.py
@@ -139,6 +139,14 @@ class TestApplyVerdictsToRun:
         assert reloaded[0].verdict is None
         assert reloaded[0].verified_confidence is None
 
+    def test_out_of_range_index_raises_clear_error(self, tmp_path):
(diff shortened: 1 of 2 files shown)


eladrianovalle (commented):
APPROVED
Clear bounds check turns the stale verdict-index failure into an actionable ValueError, and the regression test covers the expected path.

Must fix
None.

Consider
None.

PR #75, opened 2026-07-24
chore: drop the snapshot self-recovery block from /studio-update

eladrianovalle:
Promotes a change that was sitting **staged-but-uncommitted** in a local worktree (`reads-main`) into its own reviewable PR, rather than letting it ride along with unrelated work or get discarded during worktree cleanup.

**What it removes:** the Step 2 block added in #56 — the guidance telling the assistant that when a project's installed snapshot won't run at all (traceback / `ModuleNotFoundError` / `ImportError`), it should not debug the snapshot but instead re-run `check-install`/`update` from the upstream Studio repo, reading `source_path` from `.studio/VERSION`. Plus its CHANGELOG entry.

**Scope:** command markdown only (the assistant reads it; no Python runs it). No test pinned the text. The neighbouring Step 2 cases — the `WARNING:` snapshot case, the source-behind-remote case, and the `LOCAL EDITS` clobber preview — are untouched, and Step 2 reads cleanly without it. Suite green (824).

---

**Rationale needs your confirmation before merge.** I carried this diff over faithfully, but I did not write it and could not recover *why* the block was being removed — the branch it sat on (`feat/update-reads-default-branch`) is already merged and doesn't obviously supersede it. Two readings:

- **Intentional:** the recovery path is redundant or was judged wrong, and this should merge.
- **Abandoned experiment:** it was a scratch edit, and this should be closed — leaving #56's behavior in place.

Worth noting what merging costs: this removes the only instruction covering a snapshot too broken to run itself, so that case would go back to surfacing a raw traceback. Draft until you confirm.


eladrianovalle (commented):
APPROVED
The PR cleanly removes stale recovery guidance from the `/studio-update` command and matching changelog entry.

Must fix
None.

Consider
None.


---

PR #76, opened 2026-07-28
docs: study obra/superpowers — 6 proposed borrowings, 25 rejections

eladrianovalle:
Same exercise we ran against gstack, this time against [obra/superpowers](https://github.com/obra/superpowers) (MIT, v6.2.0, commit `3dcbd5c`) — Jesse Vincent's skills library for coding agents. Forty-three agents read the repo, then every borrowing candidate went through two adversarial passes: does Studio already cover this, and does it fit the architecture at the stated cost?

**Six survived. Nine were rejected after vetting, sixteen before it.** Every survivor came back smaller than it went in. The rejections are the more useful half of the doc.

## What they do better

They measure prose; we argue about it. Skill text is treated as code under test — run the scenario without the guidance, harvest the agent's real excuses, write only enough text to kill those excuses, re-run against a mandatory no-guidance control. They also classify a failure *before* choosing the form of the fix, because a prohibition that fixes a discipline slip measurably makes wrong-shaped output worse.

## The sharpest finding is about our own prompts

The contrarian's cut mandate isn't dying of forgetfulness — it's dying by citation of its own guardrails. "Guard the essence" is a nuance clause bolted onto "Default to deletion," which is exactly what the doctrine we'd be adopting condemns. Three independent vetting passes flagged those same two lines from three different angles.

## Notable rejection

The behavioral eval harness — the tempting one — is rejected on construct validity. In superpowers the skill under test *is* the whole instruction one agent receives. Our contrarian mandate is injected twice into a multi-page instructions file inside a staged debate, so a one-shot subagent handed one fragment measures a configuration we never ship. And the baseline it exists to capture is one already-shipped config line away.

## Free riders

Three bugs-in-waiting the rejections turned up, each a one-liner: the forge loop tells the writer to hand-mutate by breaking *assertions* when the methodology says break the production *code* they guard; the R3 scar text exists at two sites with divergent wording and is asserted nowhere; and nothing checks that every file in `.claude/commands/` and `.claude/workflows/` is actually on the ship list — the exact class of the finding-verifier install bug that shipped.

## What's in this PR

Just the study document and an index entry. No code changes. The proposal items that change a shipped contract get their own specs before anyone builds them — the first one (`specs/writer-escalation-channel.md`) is already written on a separate branch.

Draft because the proposal wants a read before the phased plan starts. Phase 0 is a baseline capture using a config knob that already ships, and nothing else in the plan is measurable without it.

### DIFF
--- a/studio/docs/INDEX.md
+++ b/studio/docs/INDEX.md
@@ -12,6 +12,7 @@
 - **[SESSION_ANALYTICS_PLAN.md](./SESSION_ANALYTICS_PLAN.md)** - Auto session-health metrics: design + status (first slice shipped: `session.json` at finalize, ledger auto-append, `stats` session-health block; `studio outcome` + session linking still deferred)
 - **[GSTACK_COMPARISON.md](./GSTACK_COMPARISON.md)** - Study of garrytan/gstack: what Studio borrowed, why, and the (completed) roadmap
 - **[GSTACK_SCORECARD.md](./GSTACK_SCORECARD.md)** - The gstack retrospective + how we'll measure whether the borrowings improve or degrade Studio output (per-mechanism triggers, top-line metrics, exit rule)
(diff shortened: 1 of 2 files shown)


eladrianovalle (commented):
APPROVED
The study is internally scoped as a proposal document and the index update matches the added file.

Must fix
None.

Consider
None.


---

PR #77, opened 2026-07-28
spec: writer escalation channel for /forge

eladrianovalle:
A stuck `/forge` writer has no honest way to report itself. The handoff schema requires a commit SHA where tests pass, so a writer that can't get there either invents one or fails validation — and the loop aborts with `writer_failed` and no explanation.

**The change:** one optional field for the blocker, plus a short passage making explicit that stopping is allowed and faking a finish is not. The stuck writer commits partial work under a marked label so the dead end is inspectable. No control-flow change, no config knob, no new dependency.

**Two advocate/contrarian passes, both rejected, and the rejections earned it:**

- The first cut a trigger enum whose justification was false — the precedent it cited (`why_unresolved`) shipped in #65 and is read by neither `stats.py` nor `session.py`. It also cut an entry-gate clause that was redundant against honest writers, useless against lying ones, and actively regressive for a green unit filing an advisory note.
- The second caught that the nothing-to-commit fallback would point `writer_sha` at the *previous* unit's commit — which the report then invites a human to throw away — and that the prompt would contradict itself five lines apart.

Verified by execution: the plain commit form creates nothing on a clean tree, `--allow-empty` works on both, and `^writer:` does not match `writer(stuck):` so the two tallies stay separate.

**Honest about the evidence:** no writer has ever reported anything but success in eleven runs. That is why this is one field and some prose.

### DIFF
--- a/specs/writer-escalation-channel.md
+++ b/specs/writer-escalation-channel.md
@@ -1,0 +1,295 @@
+---
+feature: Writer Escalation Channel
+slug: writer-escalation-channel
+ticket: none
+status: draft
+studio_run: .studio/output/tech/run_tech_20260728_155545
+---
+
+# Writer Escalation Channel — Architecture Spec
+
+## In Plain Language
+
+When `/forge` builds something, a writer agent does the work and an editor agent reviews it. The
+writer's only way to report back is a structured record it has to fill in completely — including the
+ID of a commit where the tests pass.
+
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
APPROVED
The writer escalation spec keeps the change deliberately small and preserves the existing gate behavior while making the stuck-writer path explicit.

Must fix
None.

Consider
None.


---

PR #78, opened 2026-07-28
spec: named-risk valve for the /forge editor's read scope

eladrianovalle:
The editor is told "Do not wander beyond that read scope" with no exception. That leaves one class of bug unfindable: code that depends on other code without importing it. Rename a config key, change a file format, alter a CLI flag — every test passes, because nothing in the import graph noticed. The editor is the only agent positioned to catch it, and it's forbidden to look.

**The change:** replace the ban with a valve. Open one more file, but only against a risk you can name, one file per risk, and report what you checked including the clean checks. Reads widen; edits do not.

**The contrarian earned its rejection twice over.**

It killed the rationale the first draft rested on: the ban does *not* orphan the existing `out_of_unit_scope` concern category, because read scope is larger than unit scope. That false claim was about to be committed into the design of record. The true argument is narrower — direct importers only catch static-import consumers.

It also caught that the "cite, don't narrate" rule illustrated itself with a **fabricated citation**: `studio/verdict.py` is 7 lines long and the example pointed at line 88. A rule against asserting what you cannot point at, modelling a made-up pointer. Deleted, with the spec recording why nobody should add it back.

**Also cut:** three of five proposed tests (one passed before the change, one was unfalsifiable, one forbade a decision that was never forbidden), and a plan to hand-record two token numbers as the valve's first data points — the spec already calls two data points insufficient, so two more would assert precision it disclaims. A one-line provenance note replaces them.

### DIFF
--- a/specs/editor-breadth-valve.md
+++ b/specs/editor-breadth-valve.md
@@ -1,0 +1,293 @@
+---
+feature: Editor Breadth Valve
+slug: editor-breadth-valve
+ticket: none
+status: draft
+studio_run: .studio/output/tech/run_tech_20260728_165447
+---
+
+# Editor Breadth Valve — Architecture Spec
+
+## In Plain Language
+
+When `/forge` finishes writing code, a second agent reviews it. That reviewer is told which files it
+may read — the ones the change touched, plus the files that import them — and then told flatly:
+"Do not wander beyond that read scope."
+
+The ban has no exception, which leaves one class of problem unfindable. Some code depends on other
+code without importing it. Something reads a config key by name. Something parses a file format.
+Something shells out to a command. Rename that key or change that format and every test still passes,
+because nothing in the test's import graph noticed. The reviewer is the one agent positioned to catch
+it, and it is forbidden from looking.
+
+This replaces the flat ban with a valve: the reviewer may open one more file, but only to check a risk
+it can *name*, one file per risk, and it has to report what it checked — including the checks that came
+back clean. Reading wider does not let it edit wider; anything it finds out there gets written down as
+a concern rather than fixed on the spot.
+
+Second, smaller piece: the reviewer must cite a `file:line` for its claims, including the ones it would
+otherwise answer with a bare "yes." "Callers still work" is not evidence. A line number is.
+
+## Architecture at a Glance
+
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
REJECTED
The spec identifies a real reviewer-scope gap, but its own proposed verification command is currently impossible to satisfy.

Must fix
specs/editor-breadth-valve.md:261 says to verify with `grep -rn "wander\|bound the editor's reads"` returning nothing outside output directories, but this new spec itself contains `wander` at specs/editor-breadth-valve.md:15 and repeats the grep string at specs/editor-breadth-valve.md:263. As written, the check fails on the tracked spec file after the spec lands. Either exclude `specs/` / this file from the grep or rewrite the verification so it only checks the shipped prompt/docs/config surface.

Consider
None.


---

eladrianovalle (commented):
REJECTED
The follow-up push adds new specs, but the previous verification blocker is still unresolved.

Must fix
specs/editor-breadth-valve.md:261 still asks reviewers to verify with `grep -rn "wander\|bound the editor's reads"` returning nothing outside output directories, while the same tracked spec still contains `wander` at specs/editor-breadth-valve.md:15 and repeats the grep at specs/editor-breadth-valve.md:263. That command still fails against the PR as written, so either exclude `specs/` / this file from the check or rewrite the verification to target only the shipped prompt/docs/config surface.

Consider
None.


---

eladrianovalle:
Rejected. The valve was gated on a cost baseline that still does not exist, and ten real editor handoffs report nobody being blocked by the current read scope — so the trade it proposes (editor input cost for cross-cutting coverage) can't be evaluated, only guessed at. The spec itself was honest about this in its provenance note.

Not wrong, just unmeasured. Recorded as declined in `studio/docs/SUPERPOWERS_COMPARISON.md` (item 5 and the Phase 3 entry) via #81, so the study stops scheduling it, and the scorecard row stays as the test it would have to pass if revisited after a Phase 0 baseline exists.

eladrianovalle:
Correction to my earlier comment: the declined note had **not** landed. I pushed it to the spec-approvals branch just after #81 merged, so it stranded there and never reached main. It is now in #83.

PR #79, opened 2026-07-28
spec: pre-committed verification for prompt-shaped features

eladrianovalle:
Some of what Studio builds is prose — a mandate telling a critic to cut harder, a blacklist of clichés, a passage letting a stuck writer stop. Tests can prove the words are present. Nothing proves they worked, so today the write-up happens after the fact and picks the framing that flatters the result.

**The change:** pre-commit the shape of the answer before the answer is known. A pass criterion as an *iff*, a baseline that must reproduce without the feature, and an evidence file created empty at approval with its headings already written — including a mandatory "What this doesn't prove." A test fails if a spec is marked `shipped` while its evidence still says `FILL_ME`.

**The contrarian earned another rejection.**

It caught that the obvious placeholder token, `<fill>`, **parses as an HTML tag and renders as zero visible characters** — so the file whose entire job is to look unfinished would render as *complete* on GitHub. Verified with `HTMLParser`. Now `FILL_ME`.

It also caught that two of six skeleton sections were the same anti-flattery job written twice, that two tables with identical columns hid the very comparison they existed to make (now one table, two rows), and that this spec was about to **exempt itself from its own convention** — it is prompt-shaped by its own test, so it now carries a Verification section and its approval creates the first real evidence file.

And it found the reason this would have died quietly: all five existing specs describe shipped, merged features and all five still say `status: approved`. Nobody has ever advanced a spec past approval here, so the template now names the flip explicitly.

**Honest limit, recorded in the spec:** this enforces that a claim of verification is *backed*, not that verification *happened*.

### DIFF
--- a/specs/prompt-feature-verification.md
+++ b/specs/prompt-feature-verification.md
@@ -1,0 +1,387 @@
+---
+feature: Pre-committed Verification for Prompt-Shaped Features
+slug: prompt-feature-verification
+ticket: none
+status: draft
+studio_run: .studio/output/tech/run_tech_20260728_192832
+---
+
+# Pre-committed Verification for Prompt-Shaped Features — Architecture Spec
+
+## In Plain Language
+
+Some of what Studio builds is prose. A mandate telling a critic to cut harder. A blacklist of clichés
+a design agent must avoid. A passage giving a stuck writer permission to stop. For features like
+these, tests can prove the words are present in the file. Nothing can prove the words *worked*.
+
+Today that gap gets filled after the fact. Someone builds the prompt change, reads a couple of runs,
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
APPROVED
The verification spec states the real enforcement limit clearly and the `approved` versus `shipped` distinction avoids making the skeleton fail at the moment it is correctly created.

Must fix
None.

Consider
None.


---

eladrianovalle (commented):
APPROVED
The current head only advances the branch without changing `specs/prompt-feature-verification.md` since the previously reviewed SHA, and the verification convention still states the enforcement boundary clearly.

Must fix
None.

Consider
None.

PR #80, opened 2026-07-28
spec: acceptance criteria on /forge units, from the approved spec

eladrianovalle:
`/forge`'s editor renders the one verdict that decides whether a unit passed, and it is judged against the unit's **title** — a single line of prose. `/spec`'s Build Plan calls itself "the bridge to `/forge`", but that bridge does not exist in code: `/forge` cannot read a spec at all.

**The gap, stated accurately.** Every one of the eleven completed `/forge` runs was governed by an approved spec that named the unit, and not one received it. A human retyped a lossy paraphrase into free text because there was no other way to get it in. The workflow was already being followed — the wire was missing.

**The change:** checkable criteria per unit in the Build Plan, a `--spec` pointer that copies them verbatim, and per-criterion grading (`pass` / `fail` / `unverifiable`) with the evidence actually checked.

**The contrarian caught the defect that mattered.** The draft would have made `mvi_verdict` true if every criterion passed — which silently deletes *"could someone use this?"*, so a unit satisfying four narrow criteria while being unusable would pass clean. That is the exact failure MVI methodology exists to prevent, arriving through the feature meant to strengthen the verdict. Criteria are now **added** to the MVI judgment, not substituted for it.

It also reversed the feature's justification. The draft claimed most runs were ad-hoc so this would be inert; the run history says ad-hoc runs are the minority. Verified by listing the run directories.

**Also cut:** a `spec_ref` field nothing could read (it lives on the unit, not the handoff, and both handoffs forbid extra properties), a second collector helper with a duplicate test battery, and two of four build units — three were markdown-only, and the loop's entry gate is "the writer's tests passed", so those gates would have been satisfied by a suite grading nothing.

**Added:** an `unverifiable` verdict value, because a binary enum forces a fabricated verdict on criteria the editor structurally cannot check; a test-harness change so the zero-regression claim is actually testable.

### DIFF
--- a/specs/unit-acceptance-criteria.md
+++ b/specs/unit-acceptance-criteria.md
@@ -1,0 +1,412 @@
+---
+feature: Unit Acceptance Criteria from the Approved Spec
+slug: unit-acceptance-criteria
+ticket: none
+status: draft
+studio_run: .studio/output/tech/run_tech_20260728_211955
+---
+
+# Unit Acceptance Criteria from the Approved Spec — Architecture Spec
+
+## In Plain Language
+
+When `/forge` builds something, a second agent reviews it and delivers the one verdict that decides
+whether the unit passed: *if we stopped here, could someone use what we built?* Right now that verdict
+is judged against the unit's **title** — a single line of prose. Nothing else. There is no list of
+checkable statements saying what "done" means for this unit, so the loop's only binding quality
+judgment rests on one sentence.
+
+Meanwhile `/spec` produces a Build Plan listing the units to build, and its own template calls that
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
APPROVED
The acceptance-criteria spec keeps the MVI verdict intact while adding per-criterion evidence, and the cut scope avoids turning authored specs into a brittle parser contract.

Must fix
None.

Consider
None.


---

eladrianovalle (commented):
APPROVED
The current head only advances the branch without changing `specs/unit-acceptance-criteria.md` since the previously reviewed SHA, and the spec still keeps criteria additive to the MVI judgment rather than replacing it.

Must fix
None.

Consider
None.

PR #81, opened 2026-07-28
chore: record the three approved specs as approved

eladrianovalle:
Three specs were approved and merged today — the writer escalation channel (#77), pre-committed verification for prompt-shaped features (#79), and unit acceptance criteria (#80). Their frontmatter still said `status: draft`.

That field is not decoration. It is the one place a spec records whether it is the source of truth to build against, and #80's own design makes it load-bearing: a `status: draft` spec is supposed to pause `/forge` as a decision point. Left as-is, all three approved specs would have tripped that pause, and #79's spec puts the same field under test with a three-value lifecycle.

Three lines, one per spec. Nothing else changed.

### DIFF
--- a/specs/prompt-feature-verification.md
+++ b/specs/prompt-feature-verification.md
@@ -2,7 +2,7 @@
 feature: Pre-committed Verification for Prompt-Shaped Features
 slug: prompt-feature-verification
 ticket: none
-status: draft
+status: approved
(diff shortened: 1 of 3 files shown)


eladrianovalle (commented):
APPROVED
This only records the three listed specs as approved, with no code or asset changes.

Must fix
None.

Consider
None.


---

PR #82, opened 2026-07-28
feat: the /forge loop grades a unit against acceptance criteria

eladrianovalle:
Unit 1 of the approved spec `specs/unit-acceptance-criteria.md`, plus an amendment to that spec which the build itself surfaced.

Hand the loop a list of checkable statements and the editor grades each one — pass, fail, or unverifiable — with the evidence it actually checked. A run with no criteria behaves as it always did.

Built through `/forge` itself, writer then editor, both stages green. 16 JS tests and 12 in the unit's Python file; full suite 827.

## What is here

- `acceptance_criteria` on the unit, read through one shared helper so a malformed payload degrades to "no criteria" instead of injecting junk into a prompt.
- Both prompts carry the criteria; the editor is told to grade against the running code and to ignore what the handoff, the commit message, or the comments *say* the unit does.
- `criteria_verdicts` on the editor's handoff — declared but not required, because the handoff refuses unknown fields and a spec-less run has nothing to grade.
- **The exit gate reads the grades itself.** Every criterion the unit carried has to come back graded `pass`. The failing ones are named in the run log, and the verdicts ride out in the return payload.
- The MVI question survives: a unit passes only if it is usable as a complete interaction **and** every criterion passes. An earlier draft would have let four narrow criteria replace that question.

## The spec amendment, and why

The editor found the important one during its own review. The original design routed a failing criterion through `mvi_verdict`: fail a criterion, set the boolean false, and the gate notices. That trusts one agent to stay consistent across two fields it fills in itself. An editor can return `verdict: 'fail'` while claiming the unit is usable, or hand back three verdicts for five criteria — and the run would have shipped unflagged. That is worse than having no criteria at all, because it manufactures confidence instead of merely lacking it. The gate now checks the grades directly, which is the same shape as the two safety nets already in the file for red-without-revert and kept-without-commit.

The second was smaller: criterion 2 asked that *both* prompts be byte-identical with zero criteria, but the spec's own editor code block adds one sentence about the new field. The criterion contradicted the design it described, so it was reworded rather than quietly ignored.

Both amendments are in `specs/unit-acceptance-criteria.md` in this branch, per §9 — an architecture change goes back through the spec.

## Verification

Two mutations to production code, each caught by the test meant to catch it: making the criteria rule always return `true` fails three JS tests; dropping it from the gate expression fails the Python wiring test. Restored green after both.


eladrianovalle (commented):
REJECTED
The acceptance-criteria gate is wired into the prompt and return path, but it can still accept grades for the wrong criteria.

Must fix
- .claude/workflows/implementation-loop.js:298: `everyCriterionPassed` checks only verdict count and `verdict === 'pass'`; it never verifies that each returned `entry.criterion` matches the carried `criteria[index]`. A self-inconsistent editor payload with pass verdicts for duplicated, reordered, or unrelated criteria would ship unflagged, which is the same false-confidence failure this gate is meant to prevent. Require the criterion text to match the unit criterion and add a regression test.

Consider
- Focused checks passed after rerunning with explicit commands: `node --test .claude/workflows/tests/workflow-shells.test.mjs` and `cd studio && python3 -m pytest tests/test_workflow_shells.py -q`.


---

eladrianovalle (commented):
APPROVED
The follow-up fixes the prior false-confidence gap by matching each carried criterion to a passing verdict with evidence, and the focused shell tests cover the changed gate.

Must fix
- None.

Consider
- Focused checks passed: `node --test .claude/workflows/tests/workflow-shells.test.mjs` and `cd studio && python3 -m pytest tests/test_workflow_shells.py -q`.


---

PR #83, opened 2026-07-28
docs: record the breadth valve as declined

eladrianovalle:
Rescued from a branch that got merged out from under it — this commit was pushed to the spec-approvals branch just after #81 merged, so it never reached main. My close comment on #78 claimed it had landed; it hadn't. This is that fix.

The study doc proposed the `/forge` editor read-scope valve as a surviving borrowing and scheduled it as Phase 3. It was specced (#78) and rejected, on the condition the doc had already named: the valve trades editor input cost for cross-cutting coverage, and neither side is measured — the per-unit editor cost was never captured, and ten real handoffs report nobody being blocked by the current read scope.

Without this note, main keeps telling a future reader to build something we declined. Two edits: the proposal item gains a "Declined" paragraph with the reasoning, and the Phase 3 roadmap entry says so instead of scheduling it. The scorecard row stays as the test it would have to pass if revisited once there is a baseline.


eladrianovalle (commented):
APPROVED
The docs-only change consistently records the breadth valve as declined and updates the phase plan without reopening implementation scope.

**Must fix**
- None.

**Consider**
- Not run; documentation-only diff.


---

PR #84, opened 2026-07-28
feat: a blocked /forge writer can stop honestly, and the entry gate is pinned

eladrianovalle:
Both units of the approved spec `specs/writer-escalation-channel.md`, in one PR because its build plan requires it: unit A alone would ship a prompt that contradicts the docs it was reconciled against.

Suite 829. 25 JS tests. Built through `/forge`, and both units were graded against acceptance criteria passed into the loop — 6/6 and 5/5, with the editor verifying each by mutation rather than by reading test names.

## The problem

The writer's only way to report back was a record it had to fill in completely, including the ID of a commit where the tests pass. That works when the writer succeeds and falls apart when it gets genuinely stuck: there is no such commit, so it could either invent a value and pretend it finished, or return nothing valid — in which case the run aborted with `writer_failed` and told you nothing about why.

## What is here

- One optional `stuck` field on the writer's handoff. `required` is untouched, so a normal run's payload is unchanged. Absence is the "not stuck" signal; there is no `stuck: false` state.
- Three paragraphs in the writer's prompt: that stopping is allowed and faking a finish is not, four **observable** triggers to escalate on rather than a feeling, and the mechanics.
- An escalating writer commits its partial work as `writer(stuck): <unit_id>`, deliberately `--allow-empty` — the likeliest blocker is "read and read, changed nothing," which leaves a clean tree where a plain commit silently creates nothing and `writer_sha` becomes a lie.
- The escalation is announced in the transcript the moment it arrives, and `/forge` leads its report with it.
- The entry gate is now a named function with its behavior pinned for the first time, including both `static_ok` subtleties that had no coverage at all before.
- No control-flow change. An escalation fails the existing gate through existing mechanics and takes the existing deliver-flagged path.

## The live smoke, which the spec calls not skippable

Ran it in a throwaway worktree, with a genuinely impossible unit — extend a function in a module that does not exist, with instructions forbidding creating it or reimplementing it elsewhere — and deliberately no hint toward escalating, since coaching it would have tested the hint instead of the feature.

The writer escalated. All four checks pass:

- `stuck` is populated in the persisted handoff, with `mvi_claimed: false` and tests honestly red (exit code 4).
- The `writer(stuck): ledger_entry_source_repo` commit exists, and `git log --grep='^writer:'` does **not** match it, so the two tallies stay separate as the spec claimed.
- The transcript announces `Writer escalated (stopped deliberately): …` — and it appears *before* `Entry gate failed`, proving the log fires whichever way the gate falls.
- Delivered flagged with no editor pass: `delivered: true, flagged: true, editorRan: false`. The existing path, no new abort branch.

It also did something better than escalate: it reported that the requirement is already satisfied by the real ledger path, named the function and line that does it, and said the honest options are to repoint the unit or drop it. That is the failure mode this feature exists to surface, and the old code would have shown you `writer_failed`.

## One deferred concern the loop filed

The extracted gate opens with a null check on the writer that is only necessary because the call site sits above the abort handling a missing writer. Move it two lines down and that guard, plus one test assertion, become unnecessary. Verified accurate. Deferred rather than done here: it needs the spec's call-site line amended, and the payoff is removing one condition. Better folded into whatever next touches that region.


eladrianovalle (commented):
REJECTED
The criteria gate is covered by focused tests, but two paths still undermine the new honesty contract.

**Must fix**
- `.claude/workflows/implementation-loop.js:295`: `editor_enabled === false` returns `flagged: false` even when `unit.acceptance_criteria` is non-empty. That lets a `/forge --spec` run with the editor mandate off log "criteria ungraded" and still ship as clean, bypassing the new requirement that every carried criterion be confirmed by a passing verdict. Please either flag this combination or reject/prohibit it before delivery.
- `.claude/workflows/implementation-loop.js:188`: the stuck path tells the writer to run the real test command and then report `tests.passed=false`. If the suite exits 0 while the writer is blocked on scope, dependency shape, or an interface contradiction, the handoff now lies about the test result. The entry gate already fails on `mvi_claimed=false`, so `tests.passed` should track the actual exit code and `stuck` should explain the block.

**Consider**
- Focused checks passed locally: `node --test .claude/workflows/tests/workflow-shells.test.mjs` and `python3 -m pytest studio/tests/test_workflow_shells.py -q`.


---

eladrianovalle (commented):
APPROVED
The follow-up closes both prior blockers: stuck writers now report the real test result, and criteria-bearing mandate-off runs no longer ship clean while ungraded.

**Must fix**
- None.

**Consider**
- Focused checks passed: `node --test .claude/workflows/tests/workflow-shells.test.mjs` and `python3 -m pytest studio/tests/test_workflow_shells.py -q`.


---

eladrianovalle:
Both Must-fix items addressed, and both were real. Thanks — the second one in particular was a contradiction I wrote and then read past several times.

**The mandate-off path shipping clean.** Fixed, and narrowly: `flagged` is now true on that path when the unit carried acceptance criteria, and false when it didn't. So the config's existing users see no new warning unless they also asked for grading. The argument for the fix is one the spec had already made in another form — it rules that a mistyped `--spec` must stop rather than "silently downgrade to an ungraded run", and arriving there through config is the same failure with a quieter cause. The original ruling was written about the general mandate-off path and covered this combination by accident.

Worth noting for the record: that line came from the per-criterion grading work that merged as #82, not from this branch. Your review found it here because this branch is stacked on it.

**The stuck path misreporting tests.** You're right that it cannot be both honest and `passed=false`. Fixed: the prompt now asks for the real exit code with `passed` matching it, and says outright not to report a red suite to mean "I am stuck" — `stuck` is what says that. Your reasoning about the gate was already provable: there's a test that hands an escalated writer a *green* suite and asserts the gate stays shut, so nothing ever depended on the lie.

Both are amendments to approved specs, so both specs carry the amendment and the reason (§9 — architecture changes go back through the spec, not around it). I also corrected four other places that described `tests.passed=false` as part of the escalation record, including the spec's Mermaid diagram: the signal is the `writer(stuck):` label, `mvi_claimed=false`, and `stuck`.

Two tests, each verified by reverting the fix and confirming the intended test fails. Suite 831 here; the stack was rebased and re-verified on top, and #86's README counts followed to 843.

On your **Consider** note — those two focused suites are green here too, and the full suite is green at every level of the stack.

PR #85, opened 2026-07-28
feat: /forge --spec reads a unit's acceptance criteria from the approved spec

eladrianovalle:
Unit 2 of `specs/unit-acceptance-criteria.md`, and the half that makes unit 1 reachable: `/forge --spec <slug> --unit <id>` resolves an approved spec, copies that unit's acceptance criteria verbatim into the run, and the editor grades the built unit against them.

This closes a gap the `/spec` template has claimed to close since it was written. Its Build Plan section calls itself "the bridge to `/forge`" — and that bridge did not exist in code. A human read the approved spec, retyped a paraphrase of a unit into a free-text argument, and the reviewer graded the paraphrase.

Markdown only. No Python, no new CLI surface, no schema change.

## What is here

- `--spec` and `--unit`, resolved with the **same** repo-shape signal the file already used to find its config module — one detection rule in `forge.md`, not two.
- A spec that can't produce criteria for the named unit **stops before the loop**, printing which of the five reasons it was, the unit ids it did find, and the two ways forward. A typo'd pointer is an error, not a silent request for an ungraded run.
- A `status: draft` spec **pauses** as a decision point rather than hard-stopping or proceeding quietly.
- Criteria are copied **verbatim** — the spec's words are what the editor grades and quotes back.
- With no `--spec`, behavior is unchanged; the plan echo just names how many specs are available, and says nothing at all in a repo with none.
- `spec.md`'s Build Plan section becomes the real template: per-unit `unit_id`, `- [ ]` criteria, out-of-scope, and five authoring rules. Plus the rule that the Build Plan is a contract, not a summary.
- `specs/unit-acceptance-criteria.md`'s own Build Plan is re-authored in the new shape, so the spec that defines the format follows it.

## Verification

All seven acceptance criteria graded `pass` by the loop's editor, each with evidence it actually checked — including a side-by-side comparison proving the new resolution rule matches the existing one rather than merely resembling it. Suite 829, unchanged.

The loop also filed a Reviewer Concern that is fixed in this PR: inserting a new Step 1 renumbered the file and broke pointers in three other documents. Those now point at steps by name instead of number, so the next insertion can't break them. One of the three was a spec claiming `forge.md` "contains no reference to `/spec` or `specs/` at all" — which this unit made false.

**Still owed, and why this stays draft:** the spec names one real `--spec` run against the live API as this unit's exit criterion, on the grounds that schema changes here have only ever failed live, never offline. Three runs have now passed `acceptance_criteria` through the loop and come back with per-criterion verdicts and no schema rejection, so the risky part is retired — but a run driven by the `--spec` flag itself, rather than by hand, is still pending. It is the next thing I do.


eladrianovalle (commented):
REJECTED
The criteria wiring has focused coverage, but a `/forge --spec` run can still report clean without the criteria ever being graded.

**Must fix**
- `.claude/workflows/implementation-loop.js:304`: `editor_enabled === false` still returns `flagged: false` even when `unit.acceptance_criteria` is non-empty, while the log says the criteria are ungraded. A spec-backed run with the editor mandate off would bypass the new definition-of-done contract and ship as clean; flag this combination or reject it before delivery.

**Consider**
- Focused checks passed on the stacked head: `node --test .claude/workflows/tests/workflow-shells.test.mjs` and `python3 -m pytest studio/tests/test_workflow_shells.py studio/tests/test_spec_verification.py -q`.


---

eladrianovalle (commented):
REJECTED
The mandate-off code fix is correct and the focused tests pass, but the spec still preserves the old rejected decision in its Key Decisions table.

**Must fix**
- `specs/unit-acceptance-criteria.md:368`: the table still says ``mandate = "off"`` with criteria ungraded ships `flagged: false` and answers "Yes, with a log clause", while the amended prose above says criteria-bearing runs ship `flagged: true` and `.claude/workflows/implementation-loop.js` now returns `flagged: promisedGrades`. Update this row so the spec's decision record matches the new contract.

**Consider**
- Focused checks passed on the current head: `node --test .claude/workflows/tests/workflow-shells.test.mjs` and `python3 -m pytest studio/tests/test_workflow_shells.py -q`.


---

eladrianovalle:
Addressed — with one correction to the record, because the finding was already fixed when you filed it.

**On the timing.** The review ran against this branch's pre-rebase head. `#84`'s fix for the same finding landed at `01:26:46Z`; the review was submitted at `01:21:17Z`, about five minutes earlier. The guard and its test are on this branch now (they arrived via the merge of `main`), and the test passes here. So I did not re-fix it — that would have been theatre.

**But you offered two remedies and I had only done one.** "Flag this combination **or reject it before delivery**" — the loop flagged it; nothing refused it up front. That is this branch's own territory, since it owns the `--spec` front door, so it is now added: resolving criteria with `editor_enabled: false` stops before the loop runs, naming the two ways forward (turn the mandate on, or drop `--spec`). The loop's flag stays as the structural backstop for anyone invoking the workflow directly.

The reasoning isn't new, just newly applied. This spec already rules that a mistyped `--spec` must stop rather than "silently downgrade to an ungraded run." Config is a quieter route to the same place, and a contradiction visible before any agent runs should cost nothing to discover. Recorded in the spec's behavior table, `forge.md` Step 5, and the usage doc.

**Housekeeping worth flagging:** you had merged `main` into this branch, and my first attempt to update it rebased over that merge — which would have dropped it. I caught it before pushing (my ancestry check had been reading a stale `FETCH_HEAD`; re-checking against a named remote ref showed the real tip was *not* an ancestor). Reset to your tip, replayed just my own commit, and pushed as a fast-forward. No force-push, your merge intact.

Suite 831 here, 843 on the verification branch stacked above. Both green, and the focused suites from your Consider note pass too.

eladrianovalle (commented):
REJECTED
The new command/docs text now says criteria plus `mandate = "off"` stops before the loop, but the spec's decision record still preserves the old clean-ship ruling.

**Must fix**
- `specs/unit-acceptance-criteria.md:369`: the Key Decisions row still says `mandate = "off"` with criteria ungraded ships `flagged: false` and answers "Yes, with a log clause", contradicting the amended behavior at `specs/unit-acceptance-criteria.md:128` and `specs/unit-acceptance-criteria.md:324`. Update this row so the durable decision table matches the stop/flag contract.

**Consider**
- I did not rerun the focused tests; this follow-up delta is documentation/spec-only and the blocker is a contradictory line in the same spec.


---

eladrianovalle:
Fixed, and the sweep found two more of the same kind.

**The row you cited** now states the amended contract: `/forge` stops before the loop, the loop delivers `flagged: true` if reached anyway, and a run carrying no criteria still ships unflagged. Marked as amended with the reason, so the table records the change rather than quietly replacing the old ruling.

**Two further leftovers from earlier amendments**, neither of which you flagged:

- The Mermaid label still routed a passing verdict to "the **existing** exit gate" — wording from the superseded design where `mvi_verdict` alone decided. The prose four lines below it already said the gate matches each criterion itself.
- The mandate-off section explained the loop's flag without mentioning that `/forge` refuses the combination up front, so it read as the whole answer instead of the backstop.

This is the second review to catch a Key Decisions row lagging an amendment made to the same spec's prose — the first was a row asserting "no new gate — the exit gate's code is untouched" after the gate had changed. So I stopped fixing the cited line and read the file end to end for anything still describing superseded behavior. Both of the above came out of that, and the earlier categories now check out clean.

**One gap found and deliberately left**, rather than fixed silently: unit 2's Build Plan has no acceptance criterion for the mandate-off refusal. Nothing there is *wrong* — no criterion claims a `--spec` run can proceed without a grader — but the behavior isn't graded either. Adding one would push that unit to seven criteria, breaking the "three to six per unit" rule this very spec teaches three sections up, and folding it into the existing stop criterion doesn't work because the two stops report different things. The unit is also already built, so a criterion added now has no grader to face. The behavior is covered in the behavior table, `forge.md` Step 5, and the usage doc. Say the word if you'd rather it be a criterion anyway.

Docs-only, so the suite is unchanged at 831. Pushed as a fast-forward (`19a68d2..eff89ab`), verified against a named remote ref rather than `FETCH_HEAD`; your merge commits are all still in the history.

eladrianovalle (commented):
APPROVED
The follow-up resolves the stale mandate-off decision record and keeps the surrounding spec language consistent.

**Must fix**
- None.

**Consider**
- I did not rerun the focused suites; the delta from `19a68d` to `eff89ab` is documentation-only in `specs/unit-acceptance-criteria.md`.


---

PR #86, opened 2026-07-29
feat: prompt-shaped features write down how you would know they work

eladrianovalle:
All three units of the approved spec `specs/prompt-feature-verification.md`, in one PR — because unit 1 ships a template promising a gate that only unit 2 delivers, and the loop's own editor flagged that as an overclaim if they landed apart.

Borrowed from [obra/superpowers](https://github.com/obra/superpowers) (study: `studio/docs/SUPERPOWERS_COMPARISON.md`, Phase 4, now marked done). Suite 841, up 12.

## The problem

Some features live in an agent's prompt, where no failing test can tell you they broke. For those, "it works" has always rested on whoever last read the output — and by the time anyone asks, the reasons are gone.

## What is here

**The convention.** `/spec` now asks a prompt-shaped feature for a `## Verification` section: the pass criterion written as an *iff* **before the build**, while disagreeing with it is still cheap; the baseline the old behavior gives you; and where the evidence goes. Approving the spec creates a `<slug>-eval-results.md` sibling from a skeleton whose headings — including "What this doesn't prove" — exist before any data does, so a missing answer is a visible hole rather than an omission nobody notices.

The judgment test is in the template itself: *could a failing pytest tell us this broke?* If yes, delete the section — the tests are the criterion and this would be theatre.

**The gate.** A spec's `status` gains a third value: `draft → approved → shipped`, where `shipped` is the claim that the feature is verified. Three rules in `studio/tests/test_spec_verification.py` make that claim cost evidence — an unknown status is rejected, a spec promising evidence must have its results file, and a `shipped` spec's file must have no placeholders left. Specs that promise no evidence are untouched, which is why the existing ones stay green.

**The record.** Changelog, both README counts set from a fresh collection, the docs index, the `/spec` usage row, and Phase 4 marked done in the study.

**Its first customer is itself.** This spec is prompt-shaped, so it carries its own Verification section and its own evidence file — sitting at `status: approved` with ten placeholders, because nobody has measured it yet. The stop condition it ships is what forbids calling it shipped.

## Three concerns the loop found, and what happened to each

**The gate could refuse a bad claim but not accept a good one.** The skeleton's instruction line read "Every `FILL_ME` marks data that does not exist yet" — and rule 3 counts raw occurrences of that token. So filling in all ten real placeholders and flipping to `shipped` still left CI red, over a sentence the same file tells you not to delete. Reworded in both places. Deliberately *not* fixed by teaching the checker to skip backticked occurrences, which would have handed anyone a way to green the gate by wrapping live placeholders in backticks. Verified with the real checker: unfilled gives one violation, filled gives none.

**The heading's prefix match was decorative.** Swapping `startswith` for line equality left all eleven tests green, because every spec and every synthetic case used the exact heading — so a widened `## Verification & Evidence` would have switched two rules off silently. The writer had listed the prefix match as load-bearing while nothing tested it. One synthetic case now does, and it is the only failure under line equality.

**Deleting a section is the cheat rule 3 misses** — an empty results file passes with zero placeholders. Recorded, not papered over: the skeleton's prose no longer claims the test catches this, and a fourth rule checking that a `shipped` spec still carries its required headings is in the spec's Risks as needing its own `/spec` pass.

## Verification of the units themselves

Units 1 and 2 were built through `/forge --spec`, with criteria read from the approved spec rather than retyped, and every criterion graded with evidence — including mutation checks: disabling each of the three rules in turn was observed failing a specific named test. Unit 3 was built directly, following the precedent these specs set for markdown-only units: the loop's entry gate is "the writer's tests passed," which on a pure record-keeping unit would be satisfied by a suite grading none of it.


eladrianovalle (commented):
REJECTED
The verification convention is covered by focused tests, but this stacked head still inherits a criteria-gate hole that can make graded `/forge` runs look clean without grading.

**Must fix**
- `.claude/workflows/implementation-loop.js:304`: `editor_enabled === false` still returns `flagged: false` even when `unit.acceptance_criteria` is non-empty, while the log says the criteria are ungraded. That means the approved-spec criteria added by the parent branch can be skipped under `mandate = "off"` and still produce a clean delivery result; flag this combination or reject it before delivery.

**Consider**
- Focused checks passed: `node --test .claude/workflows/tests/workflow-shells.test.mjs` and `python3 -m pytest studio/tests/test_workflow_shells.py studio/tests/test_spec_verification.py -q`.


---

eladrianovalle (commented):
REJECTED
The mandate-off code fix is correct and the focused tests pass, but the spec still preserves the old rejected decision in its Key Decisions table.

**Must fix**
- `specs/unit-acceptance-criteria.md:368`: the table still says ``mandate = "off"`` with criteria ungraded ships `flagged: false` and answers "Yes, with a log clause", while the amended prose above says criteria-bearing runs ship `flagged: true` and `.claude/workflows/implementation-loop.js` now returns `flagged: promisedGrades`. Update this row so the spec's decision record matches the new contract.

**Consider**
- Focused checks passed on the current head: `node --test .claude/workflows/tests/workflow-shells.test.mjs` and `python3 -m pytest studio/tests/test_workflow_shells.py studio/tests/test_spec_verification.py -q`.


---

eladrianovalle (commented):
REJECTED
The follow-up docs for the declined breadth valve look consistent, but this stacked head still carries the parent spec's old clean-ship decision for criteria with the editor mandate off.

**Must fix**
- `specs/unit-acceptance-criteria.md:369`: the Key Decisions row still says `mandate = "off"` with criteria ungraded ships `flagged: false` and answers "Yes, with a log clause", contradicting the amended behavior at `specs/unit-acceptance-criteria.md:128` and `specs/unit-acceptance-criteria.md:324`. Update this inherited row so the stacked spec no longer records the rejected contract.

**Consider**
- I did not rerun the focused tests; this follow-up delta is documentation/spec-only and the blocker is a contradictory line in the same spec.


---

eladrianovalle:
Already fixed upstream, and your merge of `main` is what closed it — nothing more was needed here.

This review and the one on #85 were filed at the same moment against both heads, naming the same inherited row. I fixed it on #85; that merged, so the corrected row is on `main`. This branch carried the old copy only because it was cut before the fix, which made "bring `main` in" the right answer rather than editing the row again — a second, independently-worded fix for one defect is how two documents start disagreeing.

We both merged `main` in at the same time. I diffed the two resolutions before doing anything: identical trees, so mine was redundant and I dropped it and kept yours. Nothing forced, nothing of yours rewritten.

Verified on your merge commit: the stale row is gone, the amended row is present, suite 843, JS shells 25 green, and both README counts still match a fresh collection.

Worth naming the pattern, because it has now cost three reviews: amending this spec's prose kept leaving its Key Decisions row behind. When the agent fixed the cited row it also swept the whole file and found two more of the same kind — a Mermaid label still routing to "the existing exit gate", and a section presenting the loop's flag as the whole answer rather than the backstop. Both came through this merge. The rule I'd take from it: an amendment isn't finished until the table, the diagram, and the summary agree with the prose.

eladrianovalle (commented):
APPROVED
The follow-up updates the stale mandate-off Key Decisions row so the spec now matches the amended code path and pre-loop refusal behavior.

**Must fix**
- None.

**Consider**
- Focused check passed on the current head: `python3 -m pytest studio/tests/test_spec_verification.py -q`.


---

PR #87, opened 2026-07-29
refactor: compute the entry gate below the missing-writer abort

eladrianovalle:
Closes a Reviewer Concern the `/forge` loop filed against its own build. It was deferred at the time because it needed the approved spec amended for a one-condition payoff, and I said I'd fold it into whatever next touched that region. #84 merging meant nothing was going to, so "later" was becoming "never".

## What was wrong

`passesEntryGate` opened with a `writer &&` guard for exactly one reason: the call site sat *above* the abort that handles a missing writer. The guard existed to survive a state the abort was already there to handle — and its only test could do nothing but assert that a defensive branch existed, which proves a line is present, not that anything works.

## What changed

The call moves below the abort, so `writer` is known to exist. The guard goes, and so does the assertion covering it.

`writer.tests` **keeps** its guard, which is where I departed from the concern's suggestion. The schema requires that field, so in principle it's equally dead — but a missing one would make the gate read as *passing* rather than crash, and failing open is the wrong way to fail. One guard removed because it was structurally unnecessary; one kept because its failure mode is silent.

## Verification

The ordering is now what carries the safety, so the ordering is what's tested. The loop body can't be loaded from JS, so it's a Python test asserting the gate is computed after the abort — plus an assertion that the guard really is gone, so the test can't quietly become decorative.

Checked by mutation: moving the call back above the abort fails that test, and only that test. Suite 832.

## Note for whoever merges

This touches `specs/writer-escalation-channel.md` a couple of lines above a change in #85, so a small textual conflict is possible if #85 merges first. It's two adjacent doc edits in different sections — nothing subtle.


eladrianovalle (commented):
APPROVED
The entry gate now relies on the missing-writer abort ordering instead of a redundant null guard, and the new wiring test pins that ordering.

**Must fix**
- None.

**Consider**
- Focused checks passed: `node --test .claude/workflows/tests/workflow-shells.test.mjs`; `python3 -m pytest studio/tests/test_workflow_shells.py -q`.


---

PR #88, opened 2026-07-29
feat(commands): add /handoff to save thread state across a context clear

eladrianovalle:
## What this does

Adds a `/handoff` command. It saves the **state** of whatever thread is in flight to a small note in the repo you're working in, so you can `/clear` and pick the work back up without losing where it stood.

## Why

Sessions run out of context. `/compact` writes a long auto-summary that's lossy in unpredictable ways and still consumes a chunk of the fresh window. The practical result is that work reaches roughly 90% and the last stretch — the part that only exists in the conversation — dies with it.

A handoff note carries the things a summary reliably drops:

- what's finished **and verified**, with the evidence (a PR number, a commit, a passing check)
- what's half-built right now, with exact paths, branches, and run IDs
- the single next concrete action, specific enough to start cold
- decisions already made, with reasoning, so they aren't re-argued next session
- what's blocked and on whom
- **landmines** — facts learned the hard way that would cost real time to rediscover

## Design notes

**Two modes, no flags.** `/handoff [note]` saves. `/handoff <slug>` resumes. If the argument matches an existing thread note it's a resume, otherwise it's a save.

**The command ships from Studio; the note lands in the consuming repo.** Path resolution mirrors how `/spec` already works, with one addition at the top: if the repo's `CLAUDE.md` names a home for durable notes (a docs tree, a vault), the note goes there. Otherwise `.studio/threads/` in a consuming repo, or `threads/` in the source repo.

**The load-bearing rule is verify-before-write.** A save has to actually run `git status`, check branches, and list open PRs before writing — not record what it thinks it remembers. A note that captures what was *believed* is worse than no note, because it sends the next session confidently in the wrong direction.

**State, not history.** No conversation narration. Overwrite rather than append — it's a snapshot, not a log.

## Scope

One new file: `.claude/commands/handoff.md` (~85 lines). Nothing else touched.

## Not ready for review yet

Opened as a draft, but note that draft state doesn't stick on private repos without a paid plan — if this shows as ready-for-review, that's why, not intent. I want to use it on a real session first and confirm the note it produces is actually sufficient to resume cold. Writing the command is the easy half.

### DIFF
--- a/.claude/commands/handoff.md
+++ b/.claude/commands/handoff.md
@@ -1,0 +1,83 @@
+---
+description: Save the in-flight thread to a durable note so you can /clear safely — or resume one
+argument-hint: "[a note to add] | <slug to resume>"
+---
+
+A session is about to run out of room, or has already been cleared. `/handoff` writes the *state* of the work in flight to a small note in this repo, so the thread survives the conversation and can be picked back up from something precise instead of a long auto-summary.
+
+The command lives in Studio; the note it writes belongs to whatever repo you are working in.
+
+Pick the mode from `$ARGUMENTS`:
+
+- **Save** — `/handoff [optional note]` writes (or overwrites) the handoff note for the thread in flight, then says in plain words that it is safe to `/clear`, and gives the exact command to resume.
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
APPROVED
The new `/handoff` command is narrowly scoped documentation for saving and resuming thread state, and I did not find blockers in the command instructions.

**Must fix**: None.

**Consider**: None.


---

PR #89, opened 2026-07-29
spec: rule 4 — a shipped spec's evidence file keeps its shape

eladrianovalle:
Spec only. Nothing built, suite unchanged at 844. Draft because a spec is the source of truth only once you approve it — and because the contrarian's recorded vote was **REJECTED** against the design it saw, which is not this one.

Closes the gap the verification spec's own Risks section recorded: rules 1-3 count placeholders, so you can reach green by deleting a section instead of answering it.

## The rule

At `status: shipped`, every required heading must be **present**, and each must carry **at least one line the reporter wrote themselves** — where "wrote themselves" means "is not a line the template printed", taken from the same skeleton extraction the heading agreement already performs.

`approved` is untouched: an unfilled skeleton there is correct compliance, and a rule that fired on it would fail people for following the convention.

## Why this went through `/spec` and not straight to `/forge`

The debate killed four things, and one of them was the whole premise I walked in with.

**Reshaping the skeleton to move guidance into HTML comments.** The command file ships to consuming repos in `SLASH_COMMANDS`; no test does. So invisible guidance would have travelled to exactly the repos with nothing to make it safe — in a feature that had already replaced one marker for rendering as zero visible characters. My argument for it ("the template costs less to change now than ever again") also undercounted: the skeleton lives in three tracked places, not one.

**Excluding blockquotes as "the printed question."** The one real evidence file writes its copied pass criterion *as a blockquote*. That rule would have failed the only correct file in the repo, and taught reporters to avoid the shape their example demonstrates.

**A looser second tier** for the results table and the pass criterion. Prototyped against the real files: it could not fire on them **at all**, because any printed prompt under a heading is a visible line forever. The first design's test list claimed those two cases would fail. They wouldn't have. That finding is what turned the first design from "too big" into "doesn't work".

**A short-circuit on rule 3**, whose stated benefit was measured and found untrue — the fixture it was protecting has no `## Verdict` heading, so it changed either way — and a `revision:` frontmatter key whose only consumer would be a grep, which is the dead-field mistake this repo already named once in the escalation spec.

## What it deliberately does not do

Closes deletion and emptiness, not dishonesty. "N/A" passes. Anyone willing to retype a guidance line to defeat the check could type "N/A" faster, and that's already recorded as undetectable in principle.

And it's aimed at a **human** failure, which the spec now says outright. A person who skips leaves holes — that's what a completeness check sees. An agent's characteristic failure is the opposite: it fills every blank with something plausible, and no completeness check sees that. Agent-side honesty here stays structural — an independent reviewer that can't see the first agent's reasoning, a gate that reads returned data instead of a summary boolean, claims that must quote what they're about. Writing that down stops rule 4 being mistaken for coverage it doesn't have.

## One weakness, named and guarded

Tying the exclusion to text rather than markup means rewording a guidance line stops it matching, and a stale unfilled section's leftover line would then read as a report. The guard: any section still holding a placeholder must contain nothing but printed lines. So a wording edit fails immediately, on the real file, with a message telling you to rebuild it — and the harmful window is exactly the window it covers.

## Two units in the Build Plan

`shape_rule` (the rule, the extraction, the two disarm guards) and `shape_rule_record`. Both within the template's three-to-six criteria rule.

The new section carries a visible **Not yet approved** line, removed at approval. `status` stays `approved` on purpose: it's the input to rules 2-4, so any value meaning "partly unapproved" would switch the gate off on the spec that defines the gate.

### DIFF
--- a/specs/prompt-feature-verification.md
+++ b/specs/prompt-feature-verification.md
@@ -238,6 +238,130 @@ evidence. Either fill it in with what you actually observed, or set this spec ba
 `status: approved` and stop describing the feature as working.
 ```
 
+### Rule 4 — the shape has to survive the claim
+
+> **Not yet approved.** Everything above this line is built, merged and green. This section is a
+> proposed revision awaiting a human's yes. Delete this line at approval.
+
+Rules 1-3 count placeholders, which leaves two ways to reach green without reporting anything, both
+recorded in Risks below: delete a whole section, or clear a placeholder and write nothing under it. The
+second is the easier one, because the file's own printed guidance keeps the section looking full.
+
+**Rule 4.** With a `## Verification` section and `status: shipped`, every heading in
+`_REQUIRED_HEADINGS` must be **present** in the results file, and each one must have **at least one
+line the reporter wrote themselves**.
+
+An independent `if`, not gated on rule 3. An earlier draft made it an `elif` to keep an existing
+assertion untouched; that justification was measured and found false — the fixture it was protecting
+has no `## Verdict` heading, so it changed either way. `_violations` returns a list precisely so one CI
+run tells you everything that is wrong.
+
+`approved` is untouched. A hollow skeleton at `approved` is the pre-registration the whole convention
+rests on; a rule 4 that fired there would fail correct compliance.
+
+#### "A line the reporter wrote themselves"
+
+The load-bearing question is how to tell a report from the boilerplate it was written next to. The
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
APPROVED
The spec-only update cleanly records Rule 4's intended enforcement and keeps the unapproved revision visibly separated from the already-built sections.

**Must fix**
None.

**Consider**
None.


---

PR #90, opened 2026-07-29
test: rule 4 — a shipped spec's evidence file must still carry every heading, answered

eladrianovalle:
Unit 4 of the approved spec `specs/prompt-feature-verification.md` (rule 4, merged as PR #89).

## What this closes

Rules 1-3 count `FILL_ME` placeholders, so there were two ways to reach a green suite while reporting
nothing: delete a whole section, or clear a placeholder and write nothing under it. The second was the
easier one, because the file's own printed guidance keeps the section looking full.

**Rule 4:** with a `## Verification` section and `status: shipped`, every required heading must be
present in the evidence file, and each must carry at least one line the reporter wrote themselves. The
test asks the template what "themselves" means — any line `/spec`'s skeleton printed is not a report.

`approved` is untouched. A hollow skeleton there is the pre-registration the whole convention rests on.

## Verified by hand, not by report

Four mutations run independently of the agent that built it, each reverted:

| Mutation | Result |
|---|---|
| Invert `_own_words` | 5 red — both tolerance cases and all three hollow cases |
| Rename the skeleton's title line | 3 red, clean asserts, not a collection error |
| Reword one printed guidance line | 1 red — the regenerable-from-skeleton guard |
| Neuter the missing-heading branch | 1 red — the deleted-heading case |

Against the one real evidence file: 0 violations as it stands at `approved`. Flipped to `shipped`
in memory it draws 4 — rule 3 on the placeholders, rule 4 on What happened / What this doesn't prove /
Verdict — and does **not** flag the pass criterion, whose copied lines are a blockquote. That
blockquote is why the design rejected excluding blockquotes: it would have failed the only correct
file in the repo.

Suite 844 → **851**. Ruff clean. The doc diff is 2 lines in each of two files, all preamble: the
sentence saying deletion is "on your honour" became false. `_printed_lines` proves no skeleton section
body changed.

## The editor's one cut

`_printed_lines` now takes the block `_skeleton_block` already found, so fence scoping lives in one
place instead of two.

## Reviewer concern (not a blocker)

The skeleton preamble is now duplicated by hand in two files with nothing keeping them in sync — the
same drift the heading-agreement test prevents for the heading list, left unguarded for the prose that
explains the rule. Recorded in `.studio/output/impl_loop/shape_rule/reviewer-concerns.md`; worth a
small follow-up, deliberately not smuggled into this unit.

Draft until unit 5 (`shape_rule_record`) lands the CHANGELOG, README counts and usage-doc updates.

### DIFF
--- a/.claude/commands/spec.md
+++ b/.claude/commands/spec.md
@@ -228,8 +228,8 @@ the file (Coding Principles §8). Then:
   criterion, say so under "What this doesn't prove."
 
   Every placeholder below marks data that does not exist yet. Replacing one is the act of reporting.
-  Deleting one without replacing it — dropping a row or a section — is what these pre-written headings
-  make conspicuous in review; the test only counts placeholders, so deletion is on your honour.
+  Dropping a whole section, or clearing a placeholder and writing nothing under its heading, fails the
+  suite once this spec says `status: shipped` — every heading here needs an answer you wrote yourself.
 
   ## Pass criterion (written before the build)
   > _Copy the iff from the spec's Verification section, word for word._
--- a/specs/prompt-feature-verification-eval-results.md
+++ b/specs/prompt-feature-verification-eval-results.md
@@ -5,8 +5,8 @@ below was copied from that spec before anything was measured. Don't edit it to m
 if it turned out to be the wrong criterion, say so under "What this doesn't prove."
 
 Every placeholder below marks data that does not exist yet. Replacing one is the act of reporting.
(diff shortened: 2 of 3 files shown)


eladrianovalle (commented):
APPROVED
Rule 4 closes the missing-or-empty evidence heading gap without breaking the approved skeleton tolerance; the focused verifier test passes locally.

**Must fix**
None.

**Consider**
None.


---

PR #91, opened 2026-07-29
docs: record rule 4 — four rules, and the deletion gap closed

eladrianovalle:
Unit 5 of `specs/prompt-feature-verification.md` — the record for rule 4. **Stacked on #90**; merge
that first. No behavior change, no test added or edited.

## What it corrects

| Document | Was | Now |
|---|---|---|
| the spec itself | opened rule 4 with "Not yet approved" | banner gone; "exactly three rules" → four; a Risks bullet said rules 2–3 |
| `CHANGELOG.md` | "Three rules, no more" and a closing "known gap: deleting a section is the cheat they miss" | four rules enumerated in the same voice; the gap sentence replaced by what closed it |
| `README.md` | 843, in both places, already one behind | **851**, from a fresh collection |
| `/spec` usage row | "with placeholders left in fails CI" | "…or with one of that file's headings dropped or left unanswered" |
| the superpowers study | "three rules … known gap" | four, with the deletion gap recorded as closed |

Every one is corrected where it sat. No new changelog entry arguing with the old one — that habit is
what got three of this stack's PRs sent back.

Two "three rules" mentions survive on purpose: unit 2's and unit 5's own acceptance criteria. They
record what each unit was graded on. Rewriting them would falsify the build record, and one of them is
the criterion this unit is judged by.

## Verified, not taken on report

- 851 collected, 851 passed; both README numbers match that run.
- The four-rule enumeration checked against `_violations`, not against the previous document.
- `grep -rn "on your honour"` → one hit, the spec quoting it as this unit's own criterion.
- The banner's remaining mentions are the Key Decisions row recording *why* it existed and unit 5's
  own description of removing it. Both correct.

## One out-of-unit correction, in its own commit

The scorecard row that measures this borrowing asked for the share of results files still at `<fill>`.
Nothing contains that token — the marker shipped as `FILL_ME`, which the same document records twenty
lines above. Run as written it measures zero forever, which is the exact failure this feature exists to
prevent, sitting in the instrument meant to catch it. One word, separate commit. The other `<fill>`
mentions there quote superpowers' original proposal and are untouched.

Worth a look while you're in that row: its kill criterion ("two shipped features with unfilled results
files → either enforce the stop condition or delete the tier") now partly can't happen, because CI
enforces the stop condition. I left the wording alone — reframing a kill criterion is a judgment call,
not a record update.

## Still open from #90

The skeleton preamble now exists by hand in two files with nothing keeping them in sync. One assertion
would close it. Recorded in `.studio/output/impl_loop/shape_rule/reviewer-concerns.md`, deliberately
not smuggled into a unit that says it changes no behavior.


eladrianovalle (commented):
APPROVED
The documentation now consistently describes the fourth verification rule, the updated test count, and the `FILL_ME` measurement marker.

**Must fix**
None.

**Consider**
None.


---

PR #92, opened 2026-07-30
fix(install): ship every registered command, delete the retired ones

eladrianovalle:
Two halves of one gap in how the installer decides which slash commands a project gets. Found while running `/studio-update` on Multica: the update reported "up to date" while the project was missing a command and carrying a stale one.

## The missing command

The installer copies commands from a hand-maintained list (`SLASH_COMMANDS` in `install.py`), not from what's on disk. `/handoff` was added to `.claude/commands/` without being added to that list, so `init`/`update` have never shipped it to **any** project — and `check-install` couldn't report it missing, because "missing" is measured against the same list. Same shape as the omitted `integrations` subpackage in the CHANGELOG.

Fix: register it. One line.

## The command that wouldn't leave

The list is also what the manifest is built from, so the reverse case leaked the other way. When `/studio-implement` was renamed to `/forge`, the old file dropped out of the manifest and simply **stayed on disk** in every installed project: two commands for one job, the stale one holding the pre-rename behavior, and `check-install` calling it up to date. `extra` was already computed and thrown away (`"removed": 0,  # We don't remove extra files`).

Fix: `update` now deletes the commands and workflows Studio has stopped shipping — reported as `Removed:`, previewed by `check-install` as `Retired` — and a leftover counts against `up_to_date`, without which `update` would short-circuit and never clear it.

### What it will and won't delete

- **Only files the manifest records**, i.e. ones a past install wrote. A command the project author wrote by hand was never in the manifest, so the prune cannot reach it. There's a test for exactly that, on a run that does delete something else.
- **A retired file you edited** goes through the existing clobber guard: blocked until `--force`, which then deletes it. Same semantics `--force` already had.
- **Retired files under `.studio/source/` are left alone** — deliberately out of scope. That snapshot is Studio's own tree, a leftover module there is inert, and removing files from it is a separate decision.

## Guarding the recurrence

`TestCommandRegistration` fails if any command or workflow in the repo isn't registered with the installer, or if a registered name has no file (a typo'd entry is skipped silently today). Sibling of the existing `test_installed_snapshot_imports` guard for source modules. Verified it fails on the pre-fix list, naming `handoff.md`.

## Verification

- Full suite green: **860 passed** (9 new).
- End-to-end on a throwaway repo: fresh `init` ships `handoff.md` and manifests it; planting the real `studio-implement.md` leftover (file + manifest entry) made `check-install` print `Retired (update will delete): .claude/commands/studio-implement.md`, and `update` deleted it while a hand-written `my-own.md` beside it survived.

## Noted, not touched

`_print_local_edits_preview` prefixes every clobber-preview path with `.studio/source/`, so a `.claude/` key renders as `.studio/source/.claude/commands/foo.md`. Pre-existing display bug, adjacent to this change but not part of it.


eladrianovalle (commented):
APPROVED
Installer updates now register /handoff and safely prune manifest-owned retired Claude command/workflow files, with focused coverage for the risky paths.

Must fix
None.

Consider
The blocked-update CLI message still prefixes every locally modified manifest key as .studio/source/..., including .claude/... command files; that looks pre-existing and does not block this change.

---

eladrianovalle (commented):
APPROVED
The follow-up push keeps the installer and verifier behavior coherent: the new preamble drift guard covers the duplicated skeleton text, `/handoff` is documented in the installer surface, and the touched suite is green.

Must fix
None.

Consider
None.


---

PR #93, opened 2026-07-30
chore: /unstale pass — and /handoff actually ships now

eladrianovalle:
## The one that matters

`/handoff` shipped yesterday and worked **only in this repo**. Consuming repos receive just the files named in `install.py`'s `SLASH_COMMANDS`, and `handoff.md` was never added — so pushing a Studio update would have delivered them nothing. That is fixed, and `TestShipListParity` now asserts every command and workflow on disk is one the installer actually copies. The superpowers study already named this gap as worth guarding, and `/forge` hit it once before with a missing `.js`.

**Do not merge and push updates until this lands** — that was the whole blocker.

## Docs that contradicted the code

Not lag; actual disagreement:

| Doc said | Code does |
|---|---|
| `debate_mode` defaults to `"all_roles"` | `"per_role"` — a scope omitting it got the opposite of the promise |
| `finalize --phase` is required | optional, derived from `--run-id` |
| read loop knobs via `python -m impl_loop` (6 places) | only resolves inside the studio dir; `/forge` runs it by path |
| `.studio/` overrides shallow-merge over shipped defaults | true for roles/personas only; `scopes.toml` and `implementation_loop.toml` replace the file, and omitted keys fall back to built-in defaults |
| `verifier.py`: confidence band applies "when the caller does not pass one" | PR #71 removed that parameter; the docstring 12 lines down says so |
| exit-gate comment: "edit must hold" | the gate never reads `editor.reverted` — a revert ships the writer's green version |

Also: `prepare --json` and `finalize --artifact-root` were undocumented; `API.md`'s `studio` artifact row listed only flat-mode filenames while scoped mode is the default.

## The skeleton drift, closed

The evidence-file rules paragraph lived by hand in **three** files with nothing syncing them, and one had already drifted — the spec defining the convention still described the honour system rule 4 replaced with a test. Synced, and now asserted: reword the live skeleton in `.claude/commands/spec.md` and the copies redden instead of quietly stating a rule that no longer exists.

## Instruments that could never fire

- The superpowers phase plan claimed Phase 0 "exists". It has never been run, so every row naming a before-number is unmeasured, and the `§10 prompt doctrine` row scores a section nobody ever wrote.
- Its "`implementation_loop.toml` already ships `mandate = "off"`" was **load-bearing for rejecting the eval harness** and is wrong — it ships `"contrarian"`.
- The gstack scorecard has recorded zero ratings and zero `findings.json` in the 13 days since it landed. It now says so at the top rather than reading as measurements taken.
- Phase 2 (writer escalation) and the `/forge --spec` pointer both shipped and were still written as future work.

## Verification

Suite **851 → 855**. `ruff check .` clean. 25 JS shell tests pass.

All four new assertions mutation-checked independently: dropping `handoff.md` from the ship list reddens the parity test; rewording the live skeleton reddens the sync test; restoring the stale transcription reddens it; clean baseline passes.

## Still open, deliberately not decided here

- The scorecard kill-criterion that CI now prevents from ever firing — rewriting a kill criterion is a judgment call.
- All 9 specs sit at `status: approved` though 8 describe merged features, so `specs/` gives no signal about what is built.
- A real bug: running `python studio/run_phase.py` from the repo root makes Studio treat **its own repo** as a consuming repo, because this repo has `.studio/` without a `VERSION`. Six runs since Jul 8 landed in `.studio/output/` instead of `studio/output/`, and it scaffolds a stray `docs/studio-bridge.md`. Not touched here.


eladrianovalle (commented):
APPROVED
The /unstale documentation pass and /handoff ship-list fix line up with the current source behavior, and the new parity checks cover the drift risks they target.

Must fix
None.

Consider
This overlaps with #92 on /handoff registration and install-test coverage, so whichever PR lands second may need a small rebase or conflict cleanup.

---

PR #94, opened 2026-07-30
spec: Studio should recognize its own source repo

eladrianovalle:
Approved architecture spec for `bridge-detection`. Not an implementation — this is the document the fix gets built against.

## The bug

Running `python studio/run_phase.py <cmd>` from the repo root — the invocation `CLAUDE.md` documents — makes Studio classify **its own repo** as a consuming repo. The source repo has a bare `.studio/` (holding `update.toml` and `usage.log`) with no `VERSION`, which is precisely Studio's signature for an already-scaffolded consumer.

Since 2026-07-08 that has meant: 16 run directories filed into `.studio/output/` instead of `studio/output/` (which is empty), and a stray `docs/studio-bridge.md` scaffolded into the source tree — a welcome note addressed to itself.

## Why the fix is small

One new branch in `get_artifact_root()`. It restores an invariant the rest of the module **already documents** rather than inventing one:

- `get_output_root`, `get_knowledge_log_path`, `_project_name` and `get_outcomes_ledger_path` all already branch on `artifact_root == studio_root` meaning "this is the source repo"
- `_project_name`'s docstring states it outright: *"In this tool repo the artifact root is the `studio/` dir, so use its parent"*
- `impl_loop.py:75-89` already resolves the source repo this way, with no cwd test at all

`run_phase.get_artifact_root()` is the outlier. The fix makes the two agree.

## What the debate caught that the diagnosis didn't

**The fix would have silently turned on Slack.** `studio/.studio/integrations.toml` exists carrying `[slack] enabled = true`, and becomes readable the moment the artifact root moves. Verified independently. Digests are switched off as part of this change so enabling them stays deliberate — and because that file is gitignored, the *test* pinning the resolved path is the part that lasts, not the edit.

**Both zero-code alternatives are dead, with reasons recorded so nobody reopens them:**
- *Delete the source `.studio/`* — doesn't work. Resolution falls through to the default branch, still returns cwd, and `_scaffold_external_repo` recreates the directory and the bridge doc on the next `prepare`.
- *Add `.studio/VERSION`* — actively worse. The installer reads `VERSION` as "this repo is an installed consumer," so `check-install`/`update` would treat the source repo as a consumer of itself.

## Decisions

- **P0 (user):** migrate the 16 historical runs and the run log; no legacy-path fallback in `stats`. The repo ends with one output location.
- **P1:** identify the source repo structurally — the repo whose `studio/` dir is `get_studio_root()`. No marker file.
- Insertion point settled on evidence, not preference: no `.studio/VERSION` exists anywhere in the tree, and all 17 worktrees are siblings that `_is_within` never matches.

## No Verification section

This is not prompt-shaped. A failing pytest can say exactly when it breaks, so a pass criterion in prose beside that test would be theatre — which is the judgment the template asks for.

## Named risk

The regression test has a wrong-reason trap: a test that `chdir`s into `studio/` instead of the repo root **passes today, unfixed**, because it hits the existing `cwd == studio_root` branch. The spec requires the test be shown failing before the fix.

## Build plan

Four units in dependency order — `source_repo_detection`, `artifact_history_migration`, `integrations_side_effect`, `docs_and_invocation` — each with checkable acceptance criteria for `/forge --spec source-repo-detection --unit <id>`.

### DIFF
--- a/specs/source-repo-detection.md
+++ b/specs/source-repo-detection.md
@@ -1,0 +1,242 @@
+---
+feature: Studio recognizes its own source repo
+slug: source-repo-detection
+ticket: bridge-detection
+status: approved
+studio_run: studio/output/tech/run_tech_20260730_134434
+---
+
+# Studio Recognizes Its Own Source Repo — Architecture Spec
+
+## In Plain Language
+
+Studio writes its run artifacts into different places depending on where it's being used. In a
+project that installed Studio, artifacts go into that project's `.studio/` folder. In Studio's own
+source repo, they go into `studio/output/`. To pick between them, Studio has to answer one question:
+*am I running inside my own source repo, or inside somebody's project?*
+
+It gets that question wrong. Studio decides "this is somebody's project" partly by looking for a
+`.studio/` folder — and Studio's own repo has one, because that's where a couple of its own settings
+live. So when you run Studio from its repo root, which is exactly what the README tells you to do,
+Studio concludes it's a guest in a stranger's project. It files its work in the wrong drawer and
+leaves behind a welcome note (`docs/studio-bridge.md`) addressed to a project that is actually
+itself.
+
+Nothing crashes, which is why this went unnoticed since July 8th. But sixteen runs' worth of history
+is sitting in the wrong folder, `studio/output/` is empty, and a stray file keeps reappearing. This
+fixes the question so Studio can tell its own house from a guest room, and moves the misfiled work
+back where the rest of the code already expects to find it.
+
+## Architecture at a Glance
+
+```mermaid
+flowchart TD
+    Start["get_artifact_root()"] --> Flag{"--artifact-root<br/>passed?"}
+    Flag -- yes --> RetFlag["return that path"]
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
REJECTED
The source-repo detection spec is directionally sound, but the docs unit points implementers at files that do not exist at those paths.

**Must fix**
- `specs/source-repo-detection.md:235` - The build plan says to update root-level `API.md` and `ARCHITECTURE.md`, but this repo keeps those files at `studio/docs/API.md` and `studio/docs/ARCHITECTURE.md` (README/CLAUDE already link those paths). Please name the real paths so the acceptance criteria are executable.

**Consider**
- `specs/source-repo-detection.md:210` - The migration section says "16 historical runs", while its acceptance criteria call for 5 tech runs plus 13 `impl_loop` unit dirs. If those are intentionally different categories, a short parenthetical would keep the implementation pass from treating the counts as contradictory.


---

eladrianovalle (commented):
APPROVED
The follow-up fixes the path references and clarifies the stranded-run count, so the spec is executable now.

**Must fix**
- None.

**Consider**
- None.


---

PR #96, opened 2026-07-30
docs: merge the three write-for-humans principles into one, and fold in the response rules

eladrianovalle:
`CODING_PRINCIPLES.md` is injected verbatim into every consuming repo's CLAUDE.md. It was **131 lines there — 100% of the CLAUDE.md in `miresu` and `Multica`**, which have no project context of their own. Sections 6, 7 and 8 were 36 lines all saying "write for humans" about three different surfaces, with three near-identical framing paragraphs.

**129 → 116 lines, 9 → 7 sections, and it carries four rules it didn't have before.**

## The rule inventory

The real risk in a merge like this is silently losing a directive. The suite can't catch that — only `test_install.py` touches this file and it tests the *injection mechanism*, not the content. So here is every directive that existed, mapped to where it landed. **21 of 22 survive verbatim or reworded; one was cut, deliberately, and it's called out below.**

### From §6 — docs, comments, docstrings, commits, PRs

| Directive | Where it landed |
|---|---|
| Plain language; define a term of art or pick a simpler word | Docs bullet 1, unchanged |
| Say what it does and why, not just its name | Docs bullet 3, with the `"Refuses to overwrite your local edits"` contrast intact |
| Cut the tells of machine-written prose | Docs bullet 2, unchanged |
| Match the voice already in the file | Docs bullet 4, unchanged |
| The test: could a new teammate read it once? | Docs closing line, unchanged |

### From §7 — the code itself

| Directive | Where it landed |
|---|---|
| Simple ≠ readable; never trade readability to save lines | **Code subsection lead**, kept whole — it's the reason the code half exists |
| Name things in full (`remaining_budget` over `rb`) | Code bullet 1, unchanged |
| Explicit and verbose over compact and clever; one thing per line | Merged into Code bullet 2 |
| Plain conventional form; cleverness is a cost paid again | Merged into Code bullet 2 |
| **Don't compress just to shorten** | **CUT — see below** |
| The test: could a teammate read it top to bottom? | Code closing line, unchanged |

### From §8 — the live conversation

| Directive | Where it landed |
|---|---|
| The reader hasn't opened your files | Updates subsection lead |
| Lead with what changed and why it matters | Updates bullet 1 |
| Don't drop internal handles; keep an identifier only if actionable | Updates bullet 2, with the link/command/PR carve-out intact |
| Give context *and* a recommendation when flagging a decision | Updates bullet 3, unchanged |
| Plain and a little warm beats a dense status log | Updates bullet 6, unchanged |
| The test: could the reader act without opening a file? | Updates closing line, unchanged |

### The four new rules

| New rule | Where |
|---|---|
| Short and information-dense; skip preamble and play-by-play | Folded into Updates bullet 1 |
| End a long update with `## TL;DR` | Updates bullet 5 |
| Don't narrate self-corrections | Updates bullet 4 |
| Match a file's length to its content; edit a line over appending a section | Docs bullet 5 |

## The one cut, and why I'm flagging rather than burying it

The editor removed *"Don't compress just to shorten. Saving three lines isn't worth making the next person stop and work out what they do."*

Its reasoning: the Code lead three lines above already says **"never trade readability away to save lines"** — so the *instruction* survives and the bullet only restated it. I checked, and that's accurate rather than a "felt similar" cut. What's lost is the concrete "three lines" framing that made it vivid.

**Say the word and I'll restore it** — it's one line and the section has room.

## What the editor did

It pushed back on the writer for the right reason. The writer hit the line target partly by compressing readable prose into denser lines — the exact failure this principle warns against. The editor cut the genuine duplicate above, then **un-jammed three compound bullets** back to one idea per line, including giving the new anti-bloat rule its own visible home instead of leaving it buried mid-bullet.

Result: 26 lines against the 39 it replaces, roughly flat word count while absorbing four new rules.

## Also in here

- **A new mirror test** (`TestCodingPrinciplesMirror`) — the principles live in two hand-maintained copies (`CODING_PRINCIPLES.md` and this repo's own `CLAUDE.md`) with nothing syncing them. It caught a real pre-existing drift on its first run.
- **Four dead references fixed.** The renumbering broke citations by number, including two live instructions in the `/spec` prompt pointing at "Coding Principles §8" and "Coding Principle 9" in a file that now ends at §7. Fixed by citing **by title**, so the next renumbering can't break them again. Three false statements in the superpowers study corrected too, including two using "the file still ends at §9" as a live trigger condition.

## Verification

**862 passed** (861 + the new mirror test), `ruff check .` clean. Every directive above was verified against the shipped file by hand, not taken from the writer's self-report.

No spec for this one, deliberately: there's no architecture to discover, it's one `git revert` from undone, and the question a spec would want to answer — *does this make agents behave better?* — is unmeasurable until the superpowers Phase 0 baseline exists, which it still doesn't.

### DIFF
--- a/.claude/commands/spec.md
+++ b/.claude/commands/spec.md
@@ -203,7 +203,7 @@ truly has no structure worth drawing, it probably didn't need a spec.
 
 Present the spec to the user for approval. Summarize in plain words what the architecture landed on,
 the decisions that shaped it, and anything still open — enough that they can approve without opening
-the file (Coding Principles §8). Then:
+the file (the Coding Principles section "Write for Humans"). Then:
 
 - **On approval:** set `status: approved` in the frontmatter, save the file, and confirm the path.
   If a `--ticket` was given, remind the user to link/attach the spec on that ticket so the tracker
@@ -303,4 +303,4 @@ python ".studio/source/run_phase.py" rate --run-dir {run_dir} --score {1-5} --no
   `/forge --spec` judges the built unit against them one by one. Vague criteria there become a vague
   verdict downstream.
 - **A spec precedes the build.** Once approved, this is what `/forge` and the feature
-  work build against. (See Coding Principle 9, "Spec Before Build.")
+  work build against. (See the Coding Principles section "Spec Before Build.")
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -75,15 +75,16 @@ A new PR starts as a **draft**, not ready-for-review. Mark it ready only when th
 **Your docs, your code, and your updates are all read by a person. Write for that person.**
 
 **Docs, comments, docstrings, commit messages, PR descriptions.**
-- Use plain language, and cut the tells of machine-written prose: inflated phrasing, filler, hedging, and padding add length without adding meaning. If a term of art is unavoidable, define it the first time or pick a simpler word.
+- Use plain language. If a term of art is unavoidable, define it the first time or pick a simpler word.
+- Cut the tells of machine-written prose: inflated phrasing, filler, hedging, and padding add length without adding meaning.
 - Say what a thing does and why it matters, not just its name. "Refuses to overwrite your local edits" beats "enforces the clobber-guard precondition."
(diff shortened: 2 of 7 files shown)


eladrianovalle (commented):
APPROVED
The documentation consolidation is internally consistent, and the new parity test covers the duplicated coding-principles text without broadening runtime behavior.

**Must fix**
None.

**Consider**
None.


---

PR #97, opened 2026-07-30
spec: /forge can build in a git worktree

eladrianovalle:
Architecture spec for `forge-worktree`. **`status: draft` — not approved yet**, so `/forge --spec` would pause on it rather than build against it.

## The problem

Every git command the loop issues is a line of text in an agent's prompt, so it runs wherever that agent's shell is sitting — the main checkout, on whatever branch you had open. Building in a worktree currently needs a script that rewrites a private copy of the loop before every run, and a bare `git reset --hard` from the wrong place can move your main checkout, because worktrees share one object store.

## Why the obvious design was rejected

A consuming repo already solved this with `work_dir` + a `gitIn(u)` helper rendering `git -C <path>`. The contrarian rejected it, and the rejection is the valuable part:

`test_command`, `static_check` and `mutation_command` are interpolated into the prompts and **are not git commands**. Pinning git while leaving those running in the main repo is *worse than today* — the agent commits to the correct branch, then reports green tests that describe a completely different tree. There's no Python seam to pin them at either; the test command never passes through Python at all.

**The mechanism that survived: `cd` into the work directory first.** One instruction covers tests, static checks, mutation runs, and anything added later. `git -C` stays as the backstop for the git calls specifically.

## Three things cut, not added

- **A source-text firewall test** — red on arrival against two explanatory comments that quote git commands as prose.
- **Three named git builders** — nine tokens across seven sites in three commit shapes; one builder covering them needs flags, i.e. a config language nobody asked for.
- **An absolute `run_dir`** threaded through Python and JS — its fallback resolved one directory off (`STUDIO_ROOT` is `<repo>/studio`). Same artifact-root confusion `source-repo-detection` fixes, so deletion beat repair.

## The guard that nearly shipped vacuous

The one surviving Python check scanned for the literal `git reset --hard`. A correct implementation renders `${gitIn(u)} reset --hard` — so that literal would have had **zero occurrences** and the assertion would have passed over an empty set forever. Verified: two literal occurrences today, both rewritten by the fix. It scans for `reset --hard` now.

## One scope expansion, judged essential

`/forge`'s own Step 3 does branch selection and a dirty-tree check in the current directory. Against a worktree already on `impl/<unit_id>`, that either hard-fails or moves the main checkout off `main` — exactly the accident this prevents. Not creep.

## Carries a Verification section

Unlike `source-repo-detection`, this one is prompt-shaped where it matters. Tests can assert the prompt *says* to change directory; only a live run shows the agent *did* it. The criterion requires one run where all four hold: commits on the worktree's branch, main's `HEAD` untouched, main's working tree untouched, tests against the worktree's code.

**Honest limit, stated in the spec and required in the docs it mandates:** this reduces the blast radius. It cannot prevent the accident, because every git command is prose an agent may ignore, and validation is one-shot at t0 while agents run for an hour. Nothing will claim "worktree-safe."

## Why it matters beyond this repo

`Arkadium/solitaire-game` is 79 commits behind and has no `/handoff`, because its local `work_dir` customization blocks a normal update. Once this ships upstream, that customization becomes redundant and the repo updates cleanly with nothing lost.

Three units: the pinning layer, the flag plus Python validation, then retiring the pin script.

### DIFF
--- a/specs/forge-work-dir.md
+++ b/specs/forge-work-dir.md
@@ -1,0 +1,235 @@
+---
+feature: /forge can build in a git worktree
+slug: forge-work-dir
+ticket: forge-worktree
+status: draft
+studio_run: studio/output/tech/run_tech_20260730_152947
+---
+
+# /forge Can Build in a Git Worktree — Architecture Spec
+
+## In Plain Language
+
+`/forge` builds a unit by handing instructions to two agents: one writes the code and commits it,
+another reviews and either keeps or reverts. Every git command those agents run is a line of text in
+their instructions — `git commit`, `git reset --hard` — so each one executes wherever that agent's
+shell happens to be sitting. Which is the main checkout, on whatever branch you had open.
+
+That's fine when you're building on a branch in the main checkout. It falls apart the moment you
+want to build in a *worktree* — a second working copy of the same repo on its own branch. The agent
+reads "commit your work," runs it where it stands, and the commit lands on the wrong branch. Worse,
+`git reset --hard` from the wrong place can move your main checkout, because worktrees share one
+object store. That has happened here, and the current workaround is a script that rewrites a private
+copy of the loop before every run.
+
+This teaches the loop to be told where the code lives. You pass `--work-dir`, `/forge` checks it's
+really a worktree of this repo before anything starts, and both agents are told up front to change
+into it. One honest caveat, stated here so nobody over-trusts it: these are instructions to an agent,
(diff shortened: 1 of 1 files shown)


eladrianovalle (commented):
APPROVED
The worktree spec lays out a narrow flag-and-validation design, keeps the agent-obedience limit explicit, and splits the build into verifiable units.

Must fix
None.

Consider
- `specs/forge-work-dir.md:53` has a small wording slip: "a agent" -> "an agent".


---

PR #98, opened 2026-07-30
fix(paths): Studio recognizes its own source repo

eladrianovalle:
Unit 1 of the approved spec `specs/source-repo-detection.md`. Built through `/forge --spec`, so the editor graded all six acceptance criteria individually with evidence.

## The bug

Running `python studio/run_phase.py <cmd>` from the repo root — the invocation `CLAUDE.md` documents — made Studio classify **its own repo** as a consuming repo. The source repo has a bare `.studio/` (holding `update.toml`, `usage.log`) with no `VERSION`, which is exactly Studio's signature for an already-scaffolded consumer.

Since 2026-07-08 that meant run artifacts landing in `.studio/output/` instead of `studio/output/`, and a stray `docs/studio-bridge.md` scaffolded into the source tree — a welcome note addressed to itself.

## The fix

One new branch in `get_artifact_root()`, above the `.studio/VERSION` walk-up:

```python
if installed_root is None and _is_within(cwd, studio_root.parent):
    return studio_root
```

It restores an invariant the rest of the module **already documents**: `get_output_root`, `get_knowledge_log_path`, `_project_name` and `get_outcomes_ledger_path` all branch on `artifact_root == studio_root` meaning "this is the source repo," and `impl_loop.py` already resolves it this way with no cwd test. `run_phase` was the outlier.

## Criteria — all 6 pass, with evidence the editor gathered itself

The one worth highlighting is **criterion 2**, which demanded the test be *watched* failing rather than assumed to fail. The editor re-ran the demonstration independently rather than trusting the writer: reverted `run_phase.py` to `96a46c5` keeping the new tests, got **2 failed, 8 passed**, and captured stdout showing `Created bridge doc: .../TheGameStudio/docs/studio-bridge.md` and the run landing in `.studio/output`.

**Criterion 3** — the wrong-reason trap — is pinned two ways: the new test asserts `Path.cwd() != studio_root` before calling, so it cannot enter the pre-existing `cwd == studio_root` branch; and during the hand-revert the control test stayed green while the new one went red, demonstrating the branches are separate.

**Criterion 5** was verified for real, not just in a fixture: `prepare --phase tech` run from the actual repo root, then `ls -d docs` → "No such file or directory."

## Editor's cuts

Hoisted a duplicated `_installed_repo_root()` call to a single `installed_root` binding so the snapshot question is asked once, and tightened a six-line comment to five without dropping a fact.

## Two concerns, both verified, neither fixed here

**1. A nested consuming repo would now lose (needs your call).** `installed_root is None and _is_within(cwd, studio_root.parent)` claims *every* directory under the repo root. A repo like `_TheGameStudio/sandbox/my_game` with its own `.studio/VERSION`, driven by the source `run_phase.py`, would file artifacts into `studio/output/` instead of its own root — the walk-up used to find it.

The spec's Key Decision 2 **knowingly accepted this**, having verified no `.studio/VERSION` exists anywhere in the tree. But the evidence is stronger than "theoretical": **four existing tests had to re-nest their fixtures under `studio_src/`** because the flat layout put the pretend consumer inside the pretend source repo, where the new branch legitimately claims it.

Suggested fix is one added condition — also gate on `_find_installed_root_upwards(cwd) is None` — which keeps the branch above the walk-up while letting an explicitly-init'd nested repo win. That contradicts nothing in the spec's intent but does amend its Decision 2, so I left it for a decision rather than deciding it.

**2. A pre-existing dead clause.** `if cwd == studio_root or _is_within(cwd, studio_root)` — `_is_within` is `relative_to` in a try/except, and `p.relative_to(p)` succeeds. Verified: `_is_within(p, p)` is `True`, so the equality check can never be why the branch is taken. Predates this unit; worth a separate one-line cleanup.

## Verification

**868 passed** (862 → 868), `ruff check .` clean. Out of scope by design: the migration of stranded run directories, the Slack side effect, and the docs — those are units 2, 3 and 4.

### DIFF
--- a/studio/run_phase.py
+++ b/studio/run_phase.py
@@ -310,21 +310,20 @@ def get_artifact_root() -> Path:
 
     studio_root = get_studio_root().resolve()
     cwd = Path.cwd().resolve()
+    installed_root = _installed_repo_root(studio_root)
 
     if cwd == studio_root or _is_within(cwd, studio_root):
         # Running from the source tree (or from inside an installed snapshot).
         # In the source repo the artifact root IS the studio dir; in an installed
         # snapshot it's the consuming repo root, never the snapshot itself.
-        installed_root = _installed_repo_root(studio_root)
         return installed_root if installed_root is not None else studio_root
 
-    # Running from Studio's own repo root (or a subdir of it that isn't studio/ itself).
-    # The source repo is the one whose studio/ dir is the module we were loaded from, so
-    # a non-snapshot studio_root plus a cwd under its parent identifies it exactly. This
-    # has to come before the consumer heuristics below: the source repo keeps its own
-    # bare .studio/ for update.toml and usage.log, which would otherwise read as a
-    # scaffolded consuming repo.
-    if _installed_repo_root(studio_root) is None and _is_within(cwd, studio_root.parent):
+    # Running from Studio's own repo root, or a subdir of it other than studio/ itself.
+    # A studio_root that is not a snapshot, with cwd under its parent, is the source repo
+    # and nothing else. This has to come before the consumer heuristics below: the source
+    # repo keeps its own bare .studio/ for update.toml and usage.log, which those
+    # heuristics would otherwise read as a scaffolded consuming repo.
+    if installed_root is None and _is_within(cwd, studio_root.parent):
         return studio_root
 
     # The next two branches are distinct on purpose: an init'd repo is marked by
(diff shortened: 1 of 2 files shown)


eladrianovalle (commented):
APPROVED
The artifact-root change correctly keeps Studio source-repo runs inside `studio/` while preserving installed-snapshot and external-repo behavior.

Must fix
None.

Consider
None.


---

PR #99, opened 2026-07-30
docs: Phase 0 is run — and the metric it depended on never measured the thing

eladrianovalle:
Phase 0 has been the named blocker on three separate decisions since 2026-07-16 — the declined breadth valve, the cadence lab, and every scorecard row naming a before-number.

**It was never runnable as written.** That, not procrastination, is why it sat undone.

## Why it couldn't run

It asked for per-unit editor token cost and `editor_liveness`/`shrink_ratio` "over the existing forge runs." Neither exists:

- **Token cost is instrumented nowhere.** The loop contains zero references to metrics recording, and there are **zero `metrics.json`** files under any `impl_loop/` run directory.
- **`editor_liveness` never measured the forge editor.** `session._summarize_editor` takes *advocate-document* word counts from **debate** runs. The forge loop writes no session record, so it has never contributed to that metric and structurally cannot.

## The metric is broken on its own terms too

It reads **0% across all six session records** — the value the scorecard defines as "a dead cut mandate."

| Run | first | final | shrink |
|---|---|---|---|
| run_tech_20260708_164348 | 249 | 249 | 0.0 |
| run_tech_20260708_223215 | 160 | 160 | 0.0 |
| run_tech_20260716_163522 | 568 | 568 | 0.0 |
| run_tech_20260716_181851 | 773 | 773 | 0.0 |
| run_tech_20260730_134434 | 859 | 859 | 0.0 |
| run_tech_20260730_152947 | 618 | 4676 | **−6.5663** |

Five are single-iteration runs, so the first draft *is* the final draft and the ratio is 0.0 by arithmetic — it reports "nothing was cut" where there was never a second draft to cut to. The sixth is negative because scoped runs are **designed** to grow from a ~500-word alignment doc into a full depth doc; the metric scores the intended design as the worst result in the set.

So the number three decisions were waiting on would have read 0% no matter how well anything was working.

## The baseline that does exist

Recovered from the 12 forge handoffs on disk — which is the number people actually wanted:

| | |
|---|---|
| **Editor made real cuts** | **9/12 (75%)** |
| Reverted | 0/12 — the revert path is untested in the field |
| MVI verdict true | 12/12 — a gate that has never fired is either well-fed or asleep |
| Median rationale | 125 words against a 400 budget, so `output_budget` isn't binding |

I checked the three exceptions by hand rather than trusting a keyword match: all three genuinely open with **"No edits."** followed by reasoning about why nothing was worth cutting. That's a legitimate outcome, not a dead mandate.

## What this changes

**The breadth valve stays rejected, but the reason should be restated.** Not "we lack a baseline because nobody ran Phase 0" — rather **"per-unit editor cost is not instrumented, so the valve cannot be measured until it is."** That's a smaller, fixable gap, and closing it is a *build*, not a measurement.

Also corrects the three places in the study that cite `editor_liveness` as the instrument for forge-editor questions, since it has never measured that.

**862 passed**, docs-only change.

### DIFF
--- a/studio/docs/PHASE_0_BASELINE.md
+++ b/studio/docs/PHASE_0_BASELINE.md
@@ -1,0 +1,96 @@
+# Superpowers Phase 0 — the baseline, and why it could never be run as written
+
+Captured 2026-07-30 from the run data on disk.
+
+## The headline
+
+**Phase 0 as specified is not executable, and that — not procrastination — is why it has sat undone
+since 2026-07-16 while three separate decisions deferred to it.**
+
+The phase says: *"record the current per-unit editor token cost and `editor_liveness` /
+`shrink_ratio` over the existing forge runs."* Neither half works:
+
+1. **Per-unit editor token cost is recorded nowhere.** `.claude/workflows/implementation-loop.js`
+   contains zero references to `record-metrics` or a session record, and there are **0
+   `metrics.json` files** under any `impl_loop/` run directory. The instrument does not exist.
+2. **`editor_liveness` / `shrink_ratio` do not measure the forge editor.** `session._summarize_editor`
+   takes *advocate document word counts* from a debate run — first draft vs final. It is a metric
+   about the **debate contrarian's** effect on advocate docs. The forge loop writes no session record
+   at all, so it contributes nothing to it, ever.
+
+## The metric is also broken on its own terms
+
+All 6 session records in the repo:
+
+| Run | first | final | shrink_ratio |
+|---|---|---|---|
+| run_tech_20260708_164348 | 249 | 249 | 0.0 |
+| run_tech_20260708_223215 | 160 | 160 | 0.0 |
+| run_tech_20260716_163522 | 568 | 568 | 0.0 |
+| run_tech_20260716_181851 | 773 | 773 | 0.0 |
+| run_tech_20260730_134434 | 859 | 859 | 0.0 |
+| run_tech_20260730_152947 | 618 | 4676 | **−6.5663** |
+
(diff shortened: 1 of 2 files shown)


eladrianovalle (commented):
REJECTED
The Phase 0 baseline write-up is useful, but the comparison doc still contains stale "not run/no baseline" language in the same updated section.

Must fix
- `studio/docs/SUPERPOWERS_COMPARISON.md:448` still says Phase 3 was declined because the Phase 0 cost baseline "still does not exist", but this PR now says Phase 0 ran and recovered a 75% forge cut-rate baseline; update this to the narrower remaining gap that token cost is not instrumented.
- `studio/docs/SUPERPOWERS_COMPARISON.md:493` still says none of the top-line questions can be tracked until Phase 0 runs and that no baseline exists as of 2026-07-29, directly contradicting the new 2026-07-30 baseline lines just above it.

Consider
- `studio/docs/SUPERPOWERS_COMPARISON.md:427` still flows from "RUN 2026-07-30" into the old imperative "Run the no-editor baseline..." wording; rewriting that paragraph as outcome plus remaining follow-up would make the state easier to read.


---

eladrianovalle (commented):
APPROVED
The follow-up resolves the stale Phase 0 contradictions, and the new artifact-root/source-repo work is covered by focused tests.

Must fix
- None.

Consider
- `studio/docs/SUPERPOWERS_COMPARISON.md:495` now names cut rate as the instrument, but the "Hurting looks like" and trigger cells still say "Liveness flat"; renaming those cells to cut-rate language would avoid carrying the old metric forward.

PR #100, opened 2026-07-30
spec: approve forge-work-dir and open its evidence file

eladrianovalle:
Approval housekeeping for the spec merged in #97, so `/forge --spec forge-work-dir` builds against it instead of pausing on a draft.

- `status: draft` → `approved`
- Creates `specs/forge-work-dir-eval-results.md` from the skeleton, **empty**, with the pass criterion copied word for word rather than summarized

The empty-file-first ordering is the point of the convention: headings written before the outcome is known can't be arranged to flatter it. Every `FILL_ME` stays in place. While they do, nobody may call the feature working and the spec stays at `approved` — flipping to `shipped` is the claim, and `test_spec_verification.py` is what makes that claim cost evidence.

Confirmed the evidence file is tracked, not gitignored, so the evidence can actually be committed when it exists.

Spec-verification suite: 20 passed.

Opened ready rather than draft — this is the complete unit of work and I plan no further commits on this branch.

### DIFF
--- a/specs/forge-work-dir-eval-results.md
+++ b/specs/forge-work-dir-eval-results.md
@@ -1,0 +1,39 @@
+# /forge can build in a git worktree — Verification Results
+
+**Spec:** [`forge-work-dir.md`](./forge-work-dir.md). The pass criterion below was copied from that spec before
+anything was measured. Don't edit it to match what happened — if it turned out to be the wrong
+criterion, say so under "What this doesn't prove."
+
+Every placeholder below marks data that does not exist yet. Replacing one is the act of reporting.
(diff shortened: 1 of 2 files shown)


eladrianovalle (commented):
APPROVED
The spec approval and evidence template keep the verification boundary explicit without claiming the feature works before the live run is filled in.

Must fix
- None.

Consider
- None.


---

eladrianovalle (commented):
APPROVED
The source-repo artifact-root fix now distinguishes source checkout, installed snapshots, and consuming repos with focused regression coverage.

Must fix
- None.

Consider
- None.

PR #101, opened 2026-07-30
feat(forge): the loop targets a named work directory

eladrianovalle:
Unit 1 of 3 from the approved spec `specs/forge-work-dir.md`. Built through `/forge --spec`, so the editor graded all six criteria individually.

**This unit was built by a loop pinned to a worktree — because the bug it fixes bit us while building it.** An earlier attempt ran unpinned, someone (me) ran `git checkout main` mid-run, and the writer's commit landed on `main` instead of its branch. Nothing was lost, but the editor's next step would have been `git reset --hard` with HEAD on `main`.

## The design, and why the obvious version was rejected

Wrapping every git call in `git -C` is the intuitive fix. The contrarian rejected it as **worse than doing nothing**: `test_command`, `static_check` and `mutation_command` are interpolated into the same prompts and are **not git commands**. Pin only git and the agent commits to the right branch, then reports green tests describing a completely different tree.

So: **`cd` into the work directory is the mechanism; `git -C` is the backstop.** Both prompts now open with the `cd` instruction — that one line covers tests, static checks, mutation runs, and anything added later — plus a rule to **stop rather than retry a failed pinned command as bare `git`**, since a silent fallback is exactly how the wrong branch gets moved.

## Criteria — all 6 pass

The two worth highlighting, because both are wrong-reason traps the spec named in advance:

**Criterion 5 (byte-identical prompts when `work_dir` is unset).** The editor didn't just trust the fixture — it extracted the *pre-feature* source (`git show 89f3355^:…`), rendered the prompts from it, and confirmed the fixture matches shipped output rather than the new code echoing itself. That fixture is load-bearing: regenerate it from current code and the compatibility bar silently disappears.

**Criterion 6 (the `reset --hard` guard).** It scans for `reset --hard`, **not** `git reset --hard` — after the fix those render as `${gitIn(u)} reset --hard`, so the full literal would match nothing and the assertion would pass over an empty set forever. Demonstrated failing: the editor unpinned the revert line and captured the failure output.

## One more guard, added after the editor flagged the gap

The editor noticed the pinned-token **counts** only count *pinned* tokens — a new bare `git` added to a prompt tomorrow leaves those counts unchanged and sails through. That's the same failure this unit prevents, arriving through a door the tests left open.

Added a test that strips the preamble (which legitimately quotes a bare `` `git` `` in the never-fall-back rule) and asserts nothing backticked starts with `git ` unless it's `git -C`. **Mutation-checked:** injecting a new `git status` reddens it while both counts stay green.

## Honest limit

This **reduces the blast radius**; it cannot prevent the accident. Every git command is prompt text an agent may ignore, and validation is one-shot at t0 while agents run for an hour. Nothing here claims the loop is "worktree-safe."

## Verification

**870 Python + 33 JS tests pass**, ruff clean. Out of scope by design: the `--work-dir` flag, Python validation, and retiring the pin script — units 2 and 3.


eladrianovalle (commented):
APPROVED

The work_dir prompt pinning is internally consistent, preserves the no-work-dir baseline, and the focused JS/Python guards pass.

Must fix
None.

Consider
The next validation unit should be the one that rejects shell-breaking path characters before this reaches /forge.

---

PR #102, opened 2026-07-30
docs: describe where Studio actually files its own artifacts

eladrianovalle:
Unit 4 of the approved spec `specs/source-repo-detection.md`. Unit 1 (the code fix) merged as #98; this makes the docs describe what the code now does.

## What changed

`get_artifact_root()` gained a branch, so running the CLI from the Studio repo root files artifacts in `studio/output/` instead of misclassifying the repo as a consumer. The docs still described the old behavior.

`studio/docs/ARCHITECTURE.md` now enumerates **all four branches of the resolution chain, in code order**, including the reason branches 3 and 4 must not be merged — which mirrors the "Don't merge" comment in the source and is the only prose record of that constraint.

## The care that mattered here

`.studio/output/` is **still correct for a consuming repo** — only the source-repo case changed. So the risk wasn't missing a stale line, it was "fixing" a correct one. The writer's handoff lists, separately, every statement it deliberately **left** because it correctly described a consumer: `BRIDGE_COMMANDS_TEMPLATE.md:37`, `CLAUDE_CODE_USAGE.md:319`, `run-phase.md:24`, and the README/CLAUDE.md scaffolding lines. That second list is the evidence it read rather than pattern-matched.

`API.md` needed no edit despite the spec naming it — its `<studio>/output/` statements were already correct.

## The editor deleted a test, and it was right to

`test_doc_names_where_each_kind_of_repo_files_its_runs` asserted that three literal strings appear *somewhere* in a section. Its bridge-doc assertion would have passed just as happily on a sentence saying the source repo **does** scaffold itself one — so it could never catch the drift it was written to guard, while still blocking legitimate rewording. A test that pins prose to itself with no source-of-truth anchor is the one shape this repo's doc-parity suite deliberately avoids.

What survived is the one that earns its place: `test_doc_lists_every_branch_in_code_order` pairs each doc label with a real line of `get_artifact_root`, asserting both that the line still exists **and** that doc order matches code order. **Mutation-checked:** swapping the doc labels for branches 3 and 4 turns it red.

## Verification

**869 passed**, doc-parity suite 5 passed, ruff clean. Documentation only — no production Python touched.


eladrianovalle (commented):
APPROVED

The artifact-root docs now match the current resolution chain, and the parity guard locks the order without pulling behavior into generated docs.

Must fix
None.

Consider
None.

---

PR #103, opened 2026-07-30
chore: migrate the stranded run history into studio/output/

eladrianovalle:
Unit 2 of `specs/source-repo-detection.md`. Unit 1 (#98) fixed the detection; this moves the runs it had been misfiling since 2026-07-08, so `stats` can see them again.

**`stats` now reports 8 runs where it saw 3.**

## Done by hand, deliberately

Everything being moved is **gitignored** — it exists only in the main checkout, and no worktree contains it. There's also no test for the loop to gate on. `/forge` would have added ceremony, not safety, so I did it directly and verified each step.

## Two collisions the spec didn't anticipate — neither a duplicate

- **`ledger_auto_append`** had its *writer* handoff in one location and its *editor* handoff in the other, two minutes apart. One run, split across both trees. Merging the directories reconstructs it.
- **`unit_impl_loop_config`** was **two different runs sharing a unit_id** (Jun 28 and Jul 29) with identically-named files. A plain `mv` would have silently destroyed one. The older editor handoff is preserved as `impl--unit_impl_loop_config--editor.2026-06-28.json`.

**All 28 handoff files survive.** Nothing was overwritten.

## What moved

| | |
|---|---|
| Directories moved | 19 (5 tech runs + 14 forge unit dirs) |
| `usage.log` | two files merged in timestamp order, 11 entries, none dropped |
| `knowledge/run_log.md` | moved |
| `index.md` | **regenerated, not moved** — it's derived, and the spec says so |

## The spec pointers

Eight specs carried a `studio_run:` pointer into the old location. Four now point at the moved directory. The other four pointed at runs that are **absent everywhere** — deleted by retention cleanup long before this migration, so repointing them would just name a second path that doesn't exist. Those now say so plainly instead.

## Verification

**871 passed.** `git status` shows only the 8 spec frontmatter edits — which is the spec's own acceptance criterion, since everything moved is gitignored. A live `prepare` from the repo root landed in `studio/output/` and created no stray bridge doc, confirming unit 1's fix end to end.

Remaining on this spec: unit 3 (`integrations_side_effect`).

### DIFF
--- a/specs/contrarian-finding-verifier.md
+++ b/specs/contrarian-finding-verifier.md
@@ -3,7 +3,7 @@ feature: Independent Verifier for Contrarian Findings (Studio roadmap R1)
 slug: contrarian-finding-verifier
 ticket: none (roadmap item R1 — see studio/docs/GSTACK_COMPARISON.md)
 status: approved
-studio_run: .studio/output/tech/run_tech_20260716_163522
+studio_run: studio/output/tech/run_tech_20260716_163522
 ---
 
 # Independent Verifier for Contrarian Findings — Architecture Spec
--- a/specs/doc-parity-tests.md
+++ b/specs/doc-parity-tests.md
@@ -3,7 +3,7 @@ feature: Doc-parity tests (Studio roadmap R2, reframed)
 slug: doc-parity-tests
 ticket: none (roadmap item R2 — see studio/docs/GSTACK_COMPARISON.md)
 status: approved
-studio_run: .studio/output/tech/run_tech_20260716_181851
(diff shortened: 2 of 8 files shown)


eladrianovalle (commented):
APPROVED

The tracked diff cleanly updates the spec frontmatter away from the old .studio/output references, with the missing retained-history cases stated explicitly instead of pointing at stale paths.

Must fix
None.

Consider
The actual gitignored artifact migration is not reviewable through this PR, so this review only covers the tracked spec frontmatter updates.

PR #104, opened 2026-07-30
fix: the artifact-root move must not quietly switch Slack on

eladrianovalle:
Unit 3 of `specs/source-repo-detection.md` — the last unit of that spec, and the one consequence of the fix nobody would have noticed until it fired.

## What was about to happen

Moving the artifact root also moved where a source-repo run looks for `integrations.toml`. A config at the **new** path (`studio/.studio/`) became live for the first time — and this machine's copy had `[slack] enabled = true`.

The next `finalize` from the repo root would have started **posting Slack digests as a side effect of a path fix**. The contrarian caught this during the spec debate; I verified it live before writing it down.

## The fix, and why the tests matter more than it

Digests are off there now, with a comment saying why: enabling them should be a decision, not something a path change does for you.

But `studio/.studio/` is **gitignored**, so that edit is machine-local and appears in no diff. Which means the file change protects exactly one machine — **the tests are what carry the knowledge forward.**

- **The resolution is pinned.** A test asserts which `integrations.toml` a source-repo run resolves to, and that it is *not* the repo-root one it used before. Removing the source-repo branch from `get_artifact_root()` turns it red — I watched that rather than assuming it.
- **The fresh-clone case is covered.** Every clone of this repo has no `integrations.toml` at all, so that's the common path, and it must resolve to "no enabled target" rather than an error or a default-on. Asserted against the exact condition `_maybe_notify` gates on.

## The changelog entry

The path fix never got one — units 1 and 4 didn't add it. So this writes it, and names the Slack behavior change **inside** that entry rather than beside it. A second entry arguing with the first is the habit this repo keeps paying for.

It also records that the config file is gitignored and therefore machine-local, so a reader doesn't assume the `enabled = false` travels with the repo.

## Verification

**873 passed**, ruff clean. Mutation-checked as described above.

With this, `specs/source-repo-detection.md` is fully built: unit 1 (#98), unit 4 (#102), unit 2 (#103), and unit 3 here.


eladrianovalle (commented):
APPROVED
The new assertions cover the source-repo integrations config path and the absent-config no-op behavior; `python3 -m pytest tests/test_run_phase.py -q` passes locally.

**Must fix**
None.

**Consider**
CI should cover Ruff; it was not installed in this local environment.


---

PR #105, opened 2026-07-30
feat(forge): a bad --work-dir stops the run before any agent spawns

eladrianovalle:
Unit 2 of `specs/forge-work-dir.md`. Unit 1 (#101) taught the loop to target a named directory; this makes a **bad** directory stop the run, and points `/forge`'s own procedure at the work dir.

**Built inside a work_dir** — the first run using the shipped feature instead of the hand-pinned script. The pin script actually *refused* to run beforehand, because #101 removed the bare-`git` patterns it rewrites. The feature obsoleted its own workaround.

## Why validation is Python-side

The Workflow sandbox has no filesystem or process access — it can only spawn agents and build strings, so it cannot stat a path or ask git anything. Validation runs before the workflow is invoked, mirroring how `--spec` already stops on a bad pointer.

## Four refusal reasons, each named

| Reason | When |
|---|---|
| `missing` | no directory at that path |
| `not-a-worktree` | exists, but not inside a git worktree |
| `different-repo` | a worktree of some *other* repository |
| `unquotable` | the path would break out of the rendered quoting — **added after review, see below** |

## The `unquotable` case is not a nicety

The spec claimed the validator "refuses the handful of characters that break inside double quotes." It refused none. That matters more than a missing check usually would, because the path isn't handed to a subprocess — it's interpolated into `git -C "<path>"` inside **instruction text an agent then runs**. A directory named with a double quote renders as:

```
git -C "/tmp/wt"; echo PWNED; #" add -A
```

The quoted string ends early and the rest becomes its own command. I rendered that through the real `gitIn` before writing the fix rather than reasoning about it. The check runs *before* `is_dir` on purpose: such a directory can genuinely exist, and the point is to refuse the name.

## Verification the editor did itself

**Criterion 4** required a test that fails when `--path-format=absolute` is removed. The editor didn't take the test's name for it — it deleted the flag and watched two tests redden, confirming the predicted breakage: git answered a bare relative `.git` resolved against the caller's own tree instead of the repo under test.

All three original failure cases are exercised against **real `git init` / `git worktree add`**, not mocked subprocess — the claim under test is that git's plumbing answers the way we think, and a mock would only replay our assumptions back.

The `unquotable` guard is mutation-checked too: emptying the refusal set reddens its test.

## Honest limit

Criteria 5 and 6 (scriptPath and Step 3 targeting the work dir) live in `forge.md`, which is markdown an assistant reads. I can verify the instruction is present and unambiguous; I cannot verify an agent obeys it. The spec puts that behind a live-run pass criterion in `specs/forge-work-dir-eval-results.md`, still unfilled.

## Verification

**880 passed**, ruff clean. The two counts describing "three failures" were corrected in the same pass rather than left to disagree with the code.


eladrianovalle (commented):
REJECTED
The new --work-dir validation covers most bad paths, but it still accepts a valid POSIX directory name that breaks the exact double-quoted git command the agents receive.

**Must fix**
- `studio/impl_loop.py:39` should reject backslash when it can escape the closing quote rendered by `git -C "<path>"` (at minimum a trailing `\`; simplest is to include `\` in `_UNQUOTABLE`). Today a real directory named `wt\` passes `validate_work_dir`, then renders as `git -C ".../wt\" add -A`, so the shell no longer sees the intended path/command boundary and the run fails after agents spawn instead of stopping up front.

**Consider**
- None.


---

eladrianovalle (commented):
REJECTED
The follow-up adds integration-config path coverage, but it still leaves the previously reported --work-dir quoting hole open.

**Must fix**
- `studio/impl_loop.py:39` still does not include backslash in `_UNQUOTABLE`, so a valid directory name ending in `\` can pass `validate_work_dir` and escape the closing quote in generated `git -C "<path>"` commands.

**Consider**
- None.


---

eladrianovalle (commented):
APPROVED
The new push fixes the backslash quoting gap and adds focused coverage for both the work-dir validation and integration-config path behavior.

**Must fix**
- None.

**Consider**
- None.


---

PR #106, opened 2026-07-30
docs: retire the worktree workaround, and be honest about what --work-dir cannot do

eladrianovalle:
Unit 3 — the **last unit** of `specs/forge-work-dir.md`. Documentation and deletion; no behavior change.

With this, both approved specs are fully built: `forge-work-dir` (#101, #105, this) and `source-repo-detection` (#98, #102, #103, #104).

## The workaround is gone

`.scratch/forge-queue/pin_loop_to_worktree.py` is deleted, along with the handoff note calling it "still live and still mandatory."

It had already stopped working: when I went to use it earlier today it **refused to run**, because #101 removed the bare-`git` patterns it rewrites. The feature obsoleted its own workaround before this unit got to delete it.

*(`.scratch/` is gitignored, so that deletion is real but invisible in this diff — verified by `find` across both checkouts.)*

## The honest limit is the point of this unit

Every doc describing `--work-dir` now says it **reduces the blast radius**, and none claims it *prevents* the accident or makes the loop "worktree-safe."

That isn't modesty. Every git command is prompt **text** an agent may ignore, and validation is one-shot at t0 while agents run for an hour. A test can prove the prompt says `cd`; only a live run shows the agent obeyed.

Two guards, and I want to be clear about which one carries weight:
- `test_every_doc_states_the_limit` — requires every describing doc to contain "blast radius". **This is the durable one.**
- `test_no_doc_repeats_a_known_overclaim` — a five-phrase blacklist. The editor caught that four of those five were the exact wordings this unit deleted, so it verifies "these five don't come back," not what its original name promised. **Renamed to say so**, with a docstring pointing at the durable guard. A new overclaim in different words still sails through, and lengthening the list would only add more guesses.

## Verification the editor did itself

It enumerated the describing files with its own `grep` rather than trusting the test's hardcoded list, and confirmed the "validated before the loop starts" claim against `forge.md` — Step 3 runs the validator at line 119, the Workflow isn't invoked until 216. Validation really does precede any agent spawn.

It also verified the retired-script guard isn't vacuous by running its regex directly: it matches three invocation forms and correctly ignores the prose mention "pin_loop_to_worktree.py was deleted."

## What this does NOT do, deliberately

`specs/forge-work-dir.md` stays `status: approved`, and `forge-work-dir-eval-results.md` still holds **4 `FILL_ME` placeholders**. The editor raised this as a concern and it's correct: this unit shipped confident present-tense docs for a feature whose load-bearing claim — that an agent honors the `cd` — has not been observed once in a controlled run.

Filling that needs a live `--work-dir` run plus the no-flag baseline. It's deliberately out of this unit and is the next thing I'll do.

## Verification

**887 passed**, ruff clean.


eladrianovalle (commented):
APPROVED
The docs and parity test now describe `/forge --work-dir` with the right limits: useful upfront validation and reduced blast radius, without claiming hard safety that prompt text cannot enforce.

Must fix
None.

Consider
None.


---

PR #107, opened 2026-07-31
spec: forge-work-dir is shipped — evidence filled from a live run

eladrianovalle:
`specs/forge-work-dir.md` moves `approved` → `shipped`, because its evidence file is now filled from two controlled runs rather than from memory.

This is the first spec in the repo to make that flip. Until now every spec has sat at `approved` — which is exactly the "results files sit unfilled and readers learn to skim past them" failure the scorecard watches for.

## The measurement

Same disposable 14-line probe unit, built twice.

| Arm | Result |
|---|---|
| **Baseline** — no `work_dir`, from the main checkout, worktree present | Writer commit `1fd7634` landed **in the main checkout**; probe file created there and **not** in the worktree; worktree `HEAD` unmoved. Criterion **not met**, as intended. |
| **With the feature** — `work_dir` set | Only commit `9e05f0d` on the worktree's branch; main's `HEAD` identical before and after; main's tree clean with the probe absent; probe printed `worktree probe ran in: …-wt/wd`. Criterion **met**. |

All four parts were read from `git` in both checkouts and from the test's own output — **not** from the agents' reports, and not inferred from the rendered prompts, which is what the criterion explicitly refuses.

## What the limits section actually says

It's the point of the file, so the honest version rather than a summary:

- **"Every commit" was tested against exactly one commit.** The editor found nothing to cut in a 14-line probe, so it made no commit of its own. That an editor commit or a revert also lands in the worktree is **not** established here — only unit tests on rendered prompts cover those, which is the inference this criterion exists to refuse.
- **The probe measures where the *test* ran, not where git committed.** The loop's own editor raised this mid-run and it's correct. Criteria 1-3 survive because they were checked against git directly; criterion 4 rests on the probe alone and would miss a split.
- **The arms weren't identical in shape** — baseline ran writer-only to halve cost.
- **A single obedient run and a run that couldn't have disobeyed look the same from outside.** Only repetition separates them, and there has been none.
- **The most convincing baseline evidence isn't in the table:** earlier the same day an unpinned run committed onto `main` for real. That was an accident, not a controlled arm — but it's why the criterion is worded this way.

## Cleanup in the same pass

The probe and both scratch branches are deleted, not left behind — the editor pointed out that "delete once the measurement is done" is an instruction to a human, not a mechanism. Also removed: 19 stale worktrees, ~30 merged local branches, and the two branches from the earlier incident.

Kept deliberately: `spec/editor-breadth-valve` (holds the *declined* spec document — deleting the branch is the only way that text disappears) and `impl/impl-loop-config-mutation-teeth` (unmerged test hardening from before this session).

## Verification

**887 passed**, ruff clean. `test_spec_verification.py` now judges this evidence file for real, since the spec says `shipped` — rule 4 requires every heading to carry the author's own words rather than the skeleton's, and it passes.

### DIFF
--- a/specs/forge-work-dir-eval-results.md
+++ b/specs/forge-work-dir-eval-results.md
@@ -17,23 +17,53 @@ suite once this spec says `status: shipped` — every heading here needs an answ
 ## What happened
 | Condition | What was run | Times | Criterion met | Notes |
 |---|---|---|---|---|
-| Baseline (feature off) | FILL_ME | FILL_ME | FILL_ME | FILL_ME |
-| With the feature | FILL_ME | FILL_ME | FILL_ME | FILL_ME |
+| Baseline (feature off) | `worktree_probe_baseline` — a 14-line disposable probe unit built by the loop with **no** `work_dir`, from the main checkout on scratch branch `eval/baseline-no-workdir`, while the `wd` worktree existed | 1 | **no** | Writer commit `1fd7634` landed **in the main checkout** (`git rev-parse --show-toplevel` returned `/Users/orcpunk/Repos/_TheGameStudio`). The probe file was created in the main checkout and **not** in the worktree. The worktree's `HEAD` never moved — `907b847` before and after. The old behavior, reproduced deliberately. |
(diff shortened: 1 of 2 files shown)


eladrianovalle (commented):
APPROVED
The shipped-status spec update is internally consistent and clearly preserves the evaluation limits instead of overstating the result.

Must fix
None.

Consider
None.


---

PR #108, opened 2026-07-31
impl_loop: decide --work-dir by allowlist, not blocklist

eladrianovalle:
## What changed

`validate_work_dir` used a list of five banned characters to decide whether a `--work-dir` path could be safely rendered into the `git -C "<path>"` quoting inside agent prompts. That list was a guess, and two independent reviews each found a fresh way through it in the handful of lines that build that string.

It now works the other way round: a path may contain **letters, digits, and `/ . _ - ~` or a space**, and anything else is refused. A character nobody sat down and thought about is refused by default instead of waved through.

The known-dangerous characters (`"`, `\`, backtick, `$`) are still written down, but only to **explain** a refusal — never to decide one. Running both as independent gates would add zero refusals, since the allowlist already covers them, and would leave two lists free to drift apart.

## The refusal message

```
work-dir unquotable: this path holds characters the loop cannot safely quote:

    /Users/me/wt$foo"bar
                ^   ^
    '$' expands as a variable
    '"' ends the quoted string early

The loop renders this path into `git -C "<path>"` inside instruction text an agent
then runs, so anything outside the accepted set could change what that agent runs.
Accepted: letters, digits, and any of / . _ - ~ or a space. Rename the directory.
```

A backslash shows as the single character you typed. `repr()` would print `'\\'`, which matches nothing you can find in your own directory name.

## The cost, accepted up front

A legitimate path holding a `+`, a `(`, an `&` or an accent is refused too. That is the trade: fewer surprises from characters nobody anticipated, in exchange for renaming the occasional directory.

## Tests

Suite **893 passed** (up from 887), ruff clean. Six new tests, including the ones that pin the parts most likely to rot:

- a semicolon — never on the old banned list, and not in the known-bad map — is refused anyway
- the carets line up under the offending characters, compared as whole lines (a substring check passes on a caret one column too far right)
- every known-bad character is outside the allowlist, so the explanation can never become the thing deciding
- the accepted-set sentence users are shown still matches the real set

Four mutations spot-checked, all caught: a semicolon slipped into the allowlist, inverted carets, a backslash rendered via `repr()`, and the "rename the directory" line dropped.

## Docs

The reason name `unquotable` is unchanged, so the four-reason contract and its doc-parity test still hold. The three places describing the rule are corrected in place: `.claude/commands/forge.md` (the only one that enumerated characters — and which also claimed the refusal was one line; it is now several), `studio/docs/IMPLEMENTATION_LOOP_SPEC.md`, and `studio/docs/CLAUDE_CODE_USAGE.md`. The CHANGELOG's `--work-dir` entry is still unreleased, so it is edited rather than contradicted by a second entry.

## One note for the reviewer

The first version of the new test used an f-string that reuses its own quote character. That is Python 3.12+ syntax, the repo supports 3.10+, and the full suite passed locally because this machine runs 3.12. Ruff's `py310` target caught it. Worth knowing that a green local suite is not proof of version compatibility here.


eladrianovalle (commented):
APPROVED
The allowlist-based work-dir validation closes the prompt-quoted git path gap and keeps the documented failure modes covered.

Must fix
None.

Consider
None.


---

PR #109, opened 2026-08-02
forge: make the mutation check and the editor's green check mean something

eladrianovalle:
Two one-line defects the [superpowers study](../blob/main/studio/docs/SUPERPOWERS_COMPARISON.md) surfaced. Both have been live since the `/forge` loop shipped.

## 1. The mutation fallback measured nothing

When mutmut isn't installed, the writer was told:

> fall back to hand-mutating — break 2-3 critical **assertions**, confirm the tests FAIL, then restore.

Breaking an assertion fails its own test by construction. It proves the test *runs*; it says nothing about whether the test would catch a real bug. `AI_TDD_METHODOLOGY.md` has always specified breaking the production **code** the assertions guard — so the prompt has contradicted the methodology the whole time, and the no-mutmut arm of the gate was close to tautological.

It now asks for 2-3 changes to the production code the tests cover (flip a comparison, drop a guard clause, return a constant), and states why mutating an assertion instead is worthless — since that is precisely the mistake being corrected.

**`mutation_check.assertions_broken` → `mutations_introduced`** for the same reason. A field name is an instruction, and that one was quietly teaching the wrong check to anyone reading the schema.

## 2. A noisy passing suite counted as clean

"Test output pristine — new warnings are a finding" appeared nowhere in Studio. A unit whose diff introduced a deprecation warning or stray output shipped as clean, and the warning became someone else's problem several units later.

It's now in the editor's hard bounds, and routes to the existing **Reviewer Concerns** channel when the editor can't fix it in the pass — so it survives a revert instead of evaporating, which is what that channel exists for.

## The honest limit

Both changes are prompt text. The tests can prove the instruction *says* the right thing; only a live `/forge` run shows an agent acting on it. Three new tests in `TestMutationGateTeeth` pin the wording, and all three were mutation-checked (revert the fallback to assertions, rename the field back, delete the pristine bullet — each caught).

The `prompts-no-work-dir.json` fixture is a golden snapshot guarding that a run **without** `--work-dir` renders byte-identically. Its own comment says to regenerate only when a prompt change is intended; this is that case. I regenerated it through the test file's own loader, so the real code rendered it, and **its diff carries these two changes and nothing else** — worth a glance, since that fixture is the thing standing between an intended prompt edit and an accidental one.

## Provenance worth noting

The study that found these **rejected** the mechanism that surfaced them: the "I watched the test fail" handoff field was declined, because Studio's mutation rule already proves the stronger property with a real tool rather than an agent's attestation about its own past behavior. These two one-liners were the genuinely useful thing that analysis turned up, and they sat unclaimed for a month.

Suite **896 passed** (up from 893), ruff clean. `SUPERPOWERS_COMPARISON.md`'s Phase 1 is updated from "NOT STARTED" to what is now actually left: the prompt-doctrine section and the R3 scar constant.


eladrianovalle (commented):
APPROVED
The prompt, schema, docs, and fixture updates consistently redirect the no-mutmut fallback toward production-code mutations and add the editor warning-output guard.

Must fix
- None.

Consider
- None.


---

PR #69, opened 2026-07-16
feat(R6+R7): design-phase AI-slop blacklist + Goodwill Reservoir

eladrianovalle:
Roadmap **R6 + R7** together — a **design-phase critique guide**, borrowed from gstack's design skills and adapted for games. Injected into the **design phase only**.

## What's here

`studio/design_mandate.py` → `DESIGN_CRITIQUE_GUIDE` (mirrors the `CONTRARIAN_MANDATE` shape), wired into `build_instruction_doc` behind `if phase == "design"` inside the existing non-question-mode block — so it's present in design deliverable runs and absent from market/tech/studio and `--mode questions`. Added to `install.py` `SOURCE_FILES` for cross-repo installs.

**R6 — anti-slop, game-adapted:**
- An **11-item AI-slop blacklist** rewritten for games (not SaaS): the default purple→blue gradient backdrop, the reflexive Play/Options/Quit three-panel, ability icons in flat colored circles, centered-nothing-anchored HUDs, one bubbly radius on every surface, particles/bokeh filler, emoji standing in for real icons, the colored left-stripe card, hero copy that could sell any game, the cookie-cutter store page, shipping the system font as the display face.
- A **"would a designer who cares be embarrassed to ship this?"** self-gate ("designed, or generated?").
- An **anti-convergence directive** — name the safe default this screen type converges on, then deliberately don't ship it; interchangeable proposals are a finding.

**R7 — the Goodwill Reservoir:** a UX score starting at **70/100**, docked for player-hostile choices (hidden costs, format-punishing input, forced unskippable tutorials, dark-pattern monetization, ambiguous choices, sloppy surface) and credited for player-respecting ones (obvious top task, upfront costs, saved steps, graceful recovery). Ends with a **"report the score with its ledger"** rule so the number is auditable line-by-line, not a gut call.

## Verification
Full suite **805 passed** (3 tests: guide present in a design run, absent from market + tech). `ruff` clean on all touched files. Prose is deliberately concrete and game-specific — a slop guide that reads like slop would be self-defeating.

Built in an isolated worktree; reviewed before push (the guide content read in full).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #111, opened 2026-08-02
spec: record what the verification convention actually did, and claim it

eladrianovalle:
The convention for prompt-shaped features shipped on 2026-07-28. Its own evidence file has been blank ever since — five days in which the rule was in force and its own results went unrecorded. That was the strongest argument for deleting the tier, so it's worth saying what changed: **filling it in required no new experiments.** The evidence was already in the repo's history.

## What the criterion asked, and what happened

The pass criterion, committed before the build and unchanged since: does the *next* prompt-shaped spec carry a pass criterion committed before its build, and get its evidence file filled — with a non-empty "What this doesn't prove" — before anyone describes the feature as working?

That next spec was `forge-work-dir`. Checked against git rather than memory:

- Its `## Verification` section was present in the spec's **first** commit (`b4448ce`)
- Its evidence file was opened empty at approval (`bad3d45`)
- `git merge-base --is-ancestor` confirms both landed **before** the first build commit (`940de25`)
- It was filled from two controlled runs, and the status flipped to `shipped` in that same commit (`96c7ded`, PR #107) — so nothing called it working while the file was blank

Baseline: the prompt-shaped features that shipped in the two weeks before the convention (#69, #65, the contrarian editor mandate and pre-flight) carried no criterion at all, and no evidence file existed in the repo before this convention created its own.

**Verdict: criterion met, as written, on a sample of one.**

## The limits, which are the point

Named in full in the file, not softened. The three that matter:

- **One spec is the whole sample**, and the criterion as written was satisfiable by a single compliant example — a weakness in the criterion, not just the data.
- **The same author wrote both the rule and the spec that followed it, 48 hours apart.** Compliance that close to authorship measures salience, not durability. The real test is a prompt-shaped spec written months from now by someone who didn't write the rule.
- **This file is a counterexample to the convention's own reach.** It stayed blank for five days without ever breaking the rule, because `test_spec_verification.py` only refuses a spec claiming `status: shipped`, and this one sat at `approved`. Enforcement binds a *claim*, not a feature in use — so the way around it is to never claim completion. A reader should not conclude the convention enforces itself.

I deliberately did **not** fix that hole here. The obvious move is to stop letting `approved` be a resting place for a feature already in use, but that's a change to propose on its own, not a conclusion this evidence supports.

## Status flip

`prompt-feature-verification.md` goes `approved` → `shipped`. That's the claim the filled file earns, and it's what actually puts this spec under its own four rules — the file was previously exempt from the very check it introduced.

I verified those rules bite here rather than assuming: a leftover placeholder, a dropped heading, and a section reduced to nothing but the skeleton's own guidance are each caught. (A first attempt at that third check passed for the wrong reason — I'd removed only one of several original lines — so it was re-run stripping the section to skeleton.)

Suite **896 passed**, ruff clean. Two specs now say `shipped`, which is two more than the count that made this tier look like decoration.

### DIFF
--- a/specs/prompt-feature-verification-eval-results.md
+++ b/specs/prompt-feature-verification-eval-results.md
@@ -16,11 +16,15 @@ suite once this spec says `status: shipped` — every heading here needs an answ
 ## What happened
 | Condition | What was run | Times | Criterion met | Notes |
 |---|---|---|---|---|
-| Baseline (feature off) | FILL_ME | FILL_ME | FILL_ME | FILL_ME |
-| With the feature | FILL_ME | FILL_ME | FILL_ME | FILL_ME |
+| Baseline (feature off) | The prompt-shaped features that shipped in the two weeks before this convention existed: the design-phase slop blacklist + Goodwill Reservoir (#69), the gstack prompt borrowings — quote-to-promote, the named-scar anchor, the lone-critical override (#65), and the contrarian editor mandate + Open-Questions Pre-Flight | 3 PRs | **no** | None carried a pass criterion, and no evidence file existed in the repo before this convention created its own. `git log --diff-filter=A -- 'specs/*-eval-results.md'` returns exactly two files, both dated on or after 2026-07-28 |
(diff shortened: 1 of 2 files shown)


eladrianovalle (commented):
APPROVED
Records the verification-convention evidence and flips the spec to shipped; the changed file satisfies the existing spec-verification gate.

Must fix
None.

Consider
None.

---

PR #112, opened 2026-08-05
spec: say shipped on the eight specs that shipped

eladrianovalle:
Every spec in this repo said `approved` except two — including eight whose features have been built, merged, and in daily use for weeks. The status field couldn't tell you what had actually shipped.

That's not just untidy. It's the same blindness that let the verification convention's own evidence file sit blank for five days: `approved` had quietly become a resting place for finished work, so "not shipped" carried no information.

**Checked against the code, not memory.** Each of the eight was verified to exist on main before flipping: the verifier module and its workflow, the doc-parity test, `auto_pull_source`, the source-repo branch in `get_artifact_root`, `acceptance_criteria` in the loop, the `check-updates` command, the staleness check, and the writer's `stuck` field.

**None of them promises evidence** — they either predate the verification convention or are code-shaped features where the tests are the criterion — so `shipped` costs no results file here and all four enforcement rules stay satisfied. Suite **900 passed**, ruff clean.

**What this deliberately does not do:** close the enforcement hole recorded in `prompt-feature-verification-eval-results.md`. Nothing here stops a *future* spec parking at `approved` with a blank evidence file, because `test_spec_verification.py` still only refuses a spec that claims completion. That needs a design pass, not a status edit, and it's the next thing I'm picking up.

### DIFF
--- a/specs/contrarian-finding-verifier.md
+++ b/specs/contrarian-finding-verifier.md
@@ -2,7 +2,7 @@
 feature: Independent Verifier for Contrarian Findings (Studio roadmap R1)
 slug: contrarian-finding-verifier
 ticket: none (roadmap item R1 — see studio/docs/GSTACK_COMPARISON.md)
-status: approved
+status: shipped
 studio_run: studio/output/tech/run_tech_20260716_163522
 ---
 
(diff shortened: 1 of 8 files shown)


eladrianovalle (commented):
APPROVED
The PR cleanly marks the eight shipped specs as shipped without touching runtime code or assets.

**Must fix**
None.

**Consider**
None.

---

